PR #18, opened 2023-07-24
Add support for chat streaming

dehoward:
### Motivation and Context

responses will now be streamed directly from the Kernel to all users in the chat. 
### Description
- `ChatSkill.cs`:
  -  added `StreamResponseToClient()` to broadcast the streamed response to all clients.
  - changed `GetChatResponseAsync()` to use `Task.WhenAll` when extracting audience/user intent and semantic/document memories.
  - changed `GetChatContextTokenLimit()` to account for the tokens used in extracting the audience

### Preview
**regular messages:**
gif

**planner messages:**
gif2

**multiuser messages:**
gif

### Contribution Checklist

- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md#dev-scripts) raises no violations

### DIFF
--- a/webapi/CopilotChat/Hubs/MessageRelayHub.cs
+++ b/webapi/CopilotChat/Hubs/MessageRelayHub.cs
@@ -40,10 +40,11 @@ public class MessageRelayHub : Hub
     /// Sends a message to all users except the sender.
     /// </summary>
     /// <param name="chatId">The chat ID used as group id for SignalR.</param>
+    /// <param name="senderId">The user ID of the user that sent the message.</param>
     /// <param name="message">The message to send.</param>
-    public async Task SendMessageAsync(string chatId, object message)
+    public async Task SendMessageAsync(string chatId, string senderId, object message)
     {
-        await this.Clients.OthersInGroup(chatId).SendAsync(ReceiveMessageClientCall, chatId, message);
+        await this.Clients.OthersInGroup(chatId).SendAsync(ReceiveMessageClientCall, chatId, senderId, message);
     }
 
     /// <summary>
--- a/webapi/CopilotChat/Hubs/MessageRelayHub.cs
+++ b/webapi/CopilotChat/Hubs/MessageRelayHub.cs
@@ -30,7 +30,7 @@ public class MessageRelayHub : Hub
     /// TODO: Retrieve the user ID from the claims and call this method
     /// from the OnConnectedAsync method instead of the frontend.
     /// </summary>
-    /// <param name="chatId">The ChatID used as group id for SignalR.</param>
+    /// <param name="chatId">The chat ID used as group id for SignalR.</param>
     public async Task AddClientToGroupAsync(string chatId)
     {
         await this.Groups.AddToGroupAsync(this.Context.ConnectionId, chatId);
@@ -39,17 +39,18 @@ public class MessageRelayHub : Hub
     /// <summary>
     /// Sends a message to all users except the sender.
     /// </summary>
-    /// <param name="chatId">The ChatID used as group id for SignalR.</param>
+    /// <param name="chatId">The chat ID used as group id for SignalR.</param>
+    /// <param name="userId">The user ID used as group id for SignalR.</param>
     /// <param name="message">The message to send.</param>
-    public async Task SendMessageAsync(string chatId, object message)
+    public async Task SendMessageAsync(string chatId, string userId, object message)
     {
-        await this.Clients.OthersInGroup(chatId).SendAsync(ReceiveMessageClientCall, message, chatId);
+        await this.Clients.OthersInGroup(chatId).SendAsync(ReceiveMessageClientCall, chatId, userId, message);
     }
 
     /// <summary>
     /// Sends the typing state to all users except the sender.
     /// </summary>
-    /// <param name="chatId">The ChatID used as group id for SignalR.</param>
+    /// <param name="chatId">The chat ID used as group id for SignalR.</param>
     /// <param name="userId">The user ID of the user who is typing.</param>
     /// <param name="isTyping">Whether the user is typing.</param>
     /// <returns>A task that represents the asynchronous operation.</returns>
--- a/webapi/CopilotChat/Hubs/MessageRelayHub.cs
+++ b/webapi/CopilotChat/Hubs/MessageRelayHub.cs
@@ -40,11 +40,10 @@ public class MessageRelayHub : Hub
     /// Sends a message to all users except the sender.
     /// </summary>
     /// <param name="chatId">The chat ID used as group id for SignalR.</param>
-    /// <param name="userId">The user ID used as group id for SignalR.</param>
     /// <param name="message">The message to send.</param>
-    public async Task SendMessageAsync(string chatId, string userId, object message)
+    public async Task SendMessageAsync(string chatId, object message)
     {
-        await this.Clients.OthersInGroup(chatId).SendAsync(ReceiveMessageClientCall, chatId, userId, message);
+        await this.Clients.OthersInGroup(chatId).SendAsync(ReceiveMessageClientCall, chatId, message);
     }
 
     /// <summary>
--- a/webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
+++ b/webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
@@ -275,11 +275,14 @@ public class ChatSkill
             await this.UpdateChatMessageContentAsync(planJson, messageId);
         }
 
-        // Get the chat response
-        ChatMessage? chatMessage = await (chatContext.Variables.ContainsKey("userCancelledPlan")
-            ? this.SaveNewResponseAsync("I am sorry the plan did not meet your goals.", string.Empty, chatId, userId)
-            : this.GetChatResponseAsync(chatId, userId, chatContext));
+        // Save hardcoded response if user cancelled plan
+        if (chatContext.Variables.ContainsKey("userCancelledPlan")) {
+            await this.SaveNewResponseAsync("I am sorry the plan did not meet your goals.", string.Empty, chatId, userId);
+            return context;
+        }
 
+        // Get the chat response
+        var chatMessage = await this.GetChatResponseAsync(chatId, userId, chatContext);
         if (chatMessage == null)
         {
             context.Fail(chatContext.LastErrorDescription);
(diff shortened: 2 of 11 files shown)


webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -313,109 +294,112 @@ public async Task<string> ExtractAudienceAsync(SKContext context)
...
-        {
-            return string.Empty;
-        }
-
-        // 1. Extract user intent from the conversation history.
-        await this.UpdateResponseStatusOnClient(chatId, "Extracting user intent");
-        var userIntent = await this.GetUserIntentAsync(chatContext);
+        // Get the audience and user intent
+        await this.UpdateBotResponseStatusOnClient(chatId, "Extracting audience and user intent");
+        var tasks = await Task.WhenAll(this.GetAudienceAsync(chatContext), this.GetUserIntentAsync(chatContext));
```
TaoChenOSU, line 305: Is this going to create a data race with the same chatContext? i.e one of the method results in error but the other one finishes later and override the error state.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -313,109 +294,112 @@ public async Task<string> ExtractAudienceAsync(SKContext context)
...
+        await this.UpdateBotResponseStatusOnClient(chatId, "Extracting semantic and document memories");
+        var chatMemoriesTokenLimit = (int)(remainingToken * this._promptOptions.MemoriesResponseContextWeight);
         var documentContextTokenLimit = (int)(remainingToken * this._promptOptions.DocumentContextWeight);
-        var documentMemories = await this._documentMemorySkill.QueryDocumentsAsync(userIntent, chatId, documentContextTokenLimit, chatContext.Memory);
+        tasks = await Task.WhenAll<string>(
+            this._semanticChatMemorySkill.QueryMemoriesAsync(userIntent, chatId, chatMemoriesTokenLimit, chatContext.Memory),
+            this._documentMemorySkill.QueryDocumentsAsync(userIntent, chatId, documentContextTokenLimit, chatContext.Memory)
+        );
+
         if (chatContext.ErrorOccurred)
```
TaoChenOSU, line 348: This is not your code but we probably don't need this anymore as we are not passing the context to the `QueryXXXAsync` methods anymore. We need to throw and catch exceptions instead.

TaoChenOSU: Can be addressed in a separate PR.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -561,18 +545,19 @@ private async Task<ChatMessage> SaveNewMessageAsync(string message, string userI
     /// <param name="response">Response from the chat.</param>
     /// <param name="prompt">Prompt used to generate the response.</param>
     /// <param name="chatId">The chat ID</param>
+    /// <param name="userId">The user ID</param>
     /// <returns>The created chat message.</returns>
-    private async Task<ChatMessage> SaveNewResponseAsync(string response, string prompt, string chatId)
+    private async Task<ChatMessage?> SaveNewResponseAsync(string response, string prompt, string chatId, string userId)
```
TaoChenOSU, line 550: Is `?` needed?

TaoChenOSU (commented):
Have you tested the app with multiple users in the chat?

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
dehoward: yes, since this function returns `null` when there is an error.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
dehoward: reverted. the audience extraction code will no longer be async after #19

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
dehoward: fixed

dehoward:
> Have you tested the app with multiple users in the chat?

yes, added a couple of needed tweaks in just now and also added a gif in the description of a multiuser example.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
TaoChenOSU: Which code path returns a null?

glahaye (approved):

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
dehoward: ah sorry you're right I got confused with `GetChatResponseAsync`. I've removed the nullable from this fn

webapp/src/libs/services/ChatService.ts
```diff
@@ -122,7 +121,7 @@ export class ChatService extends BaseService {
             ask.variables = ask.variables ? ask.variables.concat(openApiSkillVariables) : openApiSkillVariables;
         }
 
-        const result = await this.getResponseAsync<IAskResult>(
```
teresaqhoang, line 125: This is needed for token usages (coming in later PR) so req to leave as is

webapp/src/components/chat/ChatRoom.tsx
```diff
@@ -65,7 +64,7 @@ export const ChatRoom: React.FC = () => {
 
     React.useEffect(() => {
         if (!shouldAutoScroll) return;
-        scrollToTarget(scrollTargetRef.current);
+        scrollViewTargetRef.current?.scrollTo(0, scrollViewTargetRef.current.scrollHeight);
```
teresaqhoang, line 67: nice!!

webapp/src/redux/features/message-relay/signalRMiddleware.ts
```diff
@@ -146,34 +144,29 @@ export const signalRMiddleware = (store: StoreMiddlewareAPI) => {
...
-        } as IChatMessage;
+                    : PlanState.Disabled;
+        }
 
-        store.dispatch({ type: 'conversations/updateConversationFromServer', payload: { message, chatId } });
+        store.dispatch({ type: 'conversations/addMessageToConversationFromServer', payload: { chatId, message } });
     });
 
+    hubConnection.on(
+        SignalRCallbackMethods.ReceiveMessageContent,
```
teresaqhoang, line 161: nit: use `ReceiveMessageStream` instead of `ReceiveMessageContent` to be more indicative of content coming through, MessageContent and Message can be mistaken as synonymous

teresaqhoang (changes requested):
Copy nit (tagging @matthewbolanos if you have any input on the following): 

In the first screengrab gif, I recommend replacing `Generating bot response...` instead of `Invoking AI model` when the uber prompt is being invoked. 
- This helps keep the flow well defined in laymens terms, and we technically invoke the AI model on multiple occasions. 

And then using `Generating Semantic Memories` instead of `Extracting Semantic Memories...`.
- @TaoChenOSU and I had a conversation about this before and extraction gets a bit confusing since it makes it seem like we're on the step that pulling the memories to use for the response generation, but "generating" implies that this step creates new memories based on the chat history, rather than extracting them from somewhere else.

webapp/src/components/chat/ChatInput.tsx
```diff
@@ -14,10 +14,7 @@ import { GetResponseOptions, useChat } from '../../libs/useChat';
 import { useAppDispatch, useAppSelector } from '../../redux/app/hooks';
 import { RootState } from '../../redux/app/store';
 import { addAlert } from '../../redux/features/app/appSlice';
-import {
```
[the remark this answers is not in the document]
dehoward: yes, fixed

dehoward:
> Copy nit (tagging @matthewbolanos if you have any input on the following):
> 
> In the first screengrab gif, I recommend replacing `Generating bot response...` instead of `Invoking AI model` when the uber prompt is being invoked.
> 
> * This helps keep the flow well defined in laymens terms, and we technically invoke the AI model on multiple occasions.
> 
> And then using `Generating Semantic Memories` instead of `Extracting Semantic Memories...`.
> 
> * @TaoChenOSU and I had a conversation about this before and extraction gets a bit confusing since it makes it seem like we're on the step that pulling the memories to use for the response generation, but "generating" implies that this step creates new memories based on the chat history, rather than extracting them from somewhere else.

good points. I've updated the strings to reflect this!

TaoChenOSU (dismissed):

webapi/CopilotChat/Hubs/MessageRelayHub.cs
```diff
@@ -39,17 +39,18 @@ public async Task AddClientToGroupAsync(string chatId)
     /// <summary>
     /// Sends a message to all users except the sender.
     /// </summary>
-    /// <param name="chatId">The ChatID used as group id for SignalR.</param>
+    /// <param name="chatId">The chat ID used as group id for SignalR.</param>
+    /// <param name="userId">The user ID used as group id for SignalR.</param>
```
teresaqhoang, line 43: Why is both chatId and userId user for groupId here?
`userId` make more sense as `senderId` or `sourceId` since it's overloaded with the "userId" property that already exists in the ChatMessage object (This was confusing when trying to follow the CreateBotMessageOnClient flow in ChatSkill)

```suggestion
    /// <param name="userId">The user ID of the source of the message.</param>
```

teresaqhoang (changes requested):
Missed the ChatSkill file in my first pass, but thanks for adding all the comments in the ChatAsync method! A lot easier to follow now

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -313,109 +297,120 @@ public async Task<string> ExtractAudienceAsync(SKContext context)
     /// <summary>
     /// Generate the necessary chat context to create a prompt then invoke the model to get a response.
     /// </summary>
+    /// <param name="chatId">The chat ID</param>
+    /// <param name="userId">The user ID</param>
     /// <param name="chatContext">The SKContext.</param>
-    /// <returns>A response from the model.</returns>
-    private async Task<string> GetChatResponseAsync(string chatId, SKContext chatContext)
+    /// <returns>The created chat message.</returns>
```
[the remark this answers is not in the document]
dehoward: included them both

teresaqhoang (dismissed):

alliscode (approved):

PR #26, opened 2023-07-24
Added Form Recognizer as another OCR Support type for Copilot Chat #2085

davearlin:
Semantic Kernel #2085

### Contribution Checklist

- [ ] The PR follows the [Contribution Guidelines](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md#dev-scripts) raises no violations


webapi/CopilotChat/Extensions/ServiceExtensions.cs
```diff
@@ -89,18 +90,21 @@ public static IServiceCollection AddPersistentOcrSupport(this IServiceCollection
 
         switch (ocrSupportConfig.Type)
         {
+            case OcrSupportOptions.OcrSupportType.AzureFormRecognizer:
+            {
+                services.AddSingleton<IOcrEngine>(sp => new AzureFormRecognizerOcrEngine(ocrSupportConfig.AzureFormRecognizer!.Endpoint!, new AzureKeyCredential(ocrSupportConfig.AzureFormRecognizer!.Key!)));
```
glahaye, line 95: The '!' operator implies that neither Endpoint nor Key can be null. But they can!

webapi/CopilotChat/Options/AzureFormRecognizerOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+/// Configuration options for Azure Form Recognizer OCR support.
+/// </summary>
+public sealed class AzureFormRecognizerOptions
+{
+    public const string PropertyName = "AzureFormRecognizer";
+
+    /// <summary>
+    /// The endpoint for accessing a provisioned Azure Form Recognizer instance
+    /// </summary>
+    [Required, NotEmptyOrWhitespace]
```
glahaye, line 18: Won't this explode if we leave these strings empty because we set OCR type to none?

glahaye: Ah, no.  I think you're good.

webapi/CopilotChat/Extensions/ServiceExtensions.cs
glahaye: Never mind.  Just saw your option class

webapi/Services/AzureFormRecognizerOcrEngine.cs
```diff
@@ -0,0 +1,64 @@
...
+    public FormRecognizerClient FormRecognizerClient { get; }
+
+    ///<inheritdoc/>
+    public async Task<string> ReadTextFromImageFileAsync(IFormFile imageFile)
+    {
+        await using (var ms = new MemoryStream())
+        {
+            await imageFile.CopyToAsync(ms);
+            var fileBytes = ms.ToArray();
+            await using var imgStream = new MemoryStream(fileBytes);
```
glahaye, line 38: I'm curious. Why do we need to create a second memory stream?

webapi/Services/AzureFormRecognizerOcrEngine.cs
```diff
@@ -0,0 +1,64 @@
...
+            // Get the text content
+            StringBuilder text = new();
+            foreach (FormPage page in formPages)
+            {
+                foreach (FormLine line in page.Lines)
+                {
+                    foreach (FormWord word in line.Words)
+                    {
+                        text.Append(word.Text);
+                        text.Append(' ');
```
glahaye, line 56: Consider using String.Join() instead of this inner loop. It would also get rid of the trailing space at the end of each line.

webapi/Services/AzureFormRecognizerOcrEngine.cs
davearlin: We don't!

webapi/Services/AzureFormRecognizerOcrEngine.cs
davearlin: Done, thanks for pointing out the trailing space problem there !

webapi/Services/AzureFormRecognizerOcrEngine.cs
glahaye: Isn't there one on line 34 and one on line 38?

Could we not just do

```
await using (var imgStream = new MemoryStream())
{
    await imageFile.CopyToAsync(imgStream);
    ...
```

davearlin: Yes, there were 2, the PR should've been updated yesterday to only include one.   This is the link to updated file here - https://github.com/microsoft/chat-copilot/pull/26/files#diff-d7d3889b894f3d930178908a909f051b7caf23ea8302fc08e63f06689b0677aa

webapi/Services/TesseractEngineWrapper.cs
```diff
@@ -30,8 +33,18 @@ public TesseractEngineWrapper(TesseractEngine tesseractEngine)
...
     ///<inheritdoc/>
-    public Page Process(Pix image)
+    public async Task<string> ReadTextFromImageFileAsync(IFormFile imageFile)
     {
-        return this.TesseractEngine.Process(image);
+        await using (var ms = new MemoryStream())
+        {
+            await imageFile.CopyToAsync(ms);
+            var fileBytes = ms.ToArray();
+            await using var imgStream = new MemoryStream(fileBytes);
```
glahaye, line 42: Two memory stream when one could do!

webapi/Services/AzureFormRecognizerOcrEngine.cs
glahaye: Yes, saw that now.  Closing this convo.  But there's another similar piece of code in the Tesseract engine wrapper...

webapi/Services/TesseractEngineWrapper.cs
davearlin: LOL, you caught me again .  Fixed!

glahaye (approved):

glahaye (dismissed):
Just need to synch with main latest and you're good to go

TaoChenOSU (approved):

davearlin:
> Just need to synch with main latest and you're good to go

All set

glahaye (approved):

PR #29, opened 2023-07-25
Adding support for custom plugin import

teresaqhoang:
### Motivation and Context

This PR adds support for custom plugins. This implementation follows manifest requirements set by OpenAI for custom ChatGPT plugins and is defined here: https://learn.microsoft.com/en-us/semantic-kernel/ai-orchestration/chatgpt-plugins#add-the-plugin-manifest-file 

### Description

webapi
 - Added GetPluginAuthHeaders() to ChatController.cs to parse plugin auth values from request headers. This was manually parsed using hardcoded headers before.
    - Standardized all plugin headers to `x-sk-copilot-(.*)-auth`
- Updated RegisterPlannerSkillsAsync() to accept a Dictionary of auth headers and register authenticated skills with the planner's kernel. 
- Added logic to register custom plugins from context variables, including support for loading custom plugins from manifest files with authentication support for user PAT and no auth. 

webapp
- General
    - Added .well-known* to .gitignore. 
    - Consolidated all hooks into `libs/hooks` directory. Updated imports accordingly
    - Consolidated dialog styles into shared class
- Create BaseCard component to render a generic card. 
   - Add AddPluginCard component to render a card for adding custom plugins. Card action triggers Plugin Wizard
    - Create PluginCard component to render a card for existing plugins. 
- Fix PluginGallery component to use new PluginCard and AddPluginCard components. 
- Update PluginConnector component to use new Dialog classes. 
- Remove unused code from PluginCard and PluginConnector components. 
- Create PluginWizard component to handle custom plugin import. Includes three steps: `EnterManifestStep`, `ValidateManifestStep`, and a confirmation page
- Manifest Validation:
    - Domain is a valid URL (doesn't throw when constructed using new URL)
    - Meets all the property requirements as [defined by OpenAI](https://learn.microsoft.com/en-us/semantic-kernel/ai-orchestration/chatgpt-plugins#add-the-plugin-manifest-file)
- If manifest domain is valid and user imports plugin, a new card will be added for that plugin at the end of the plugin gallery. User must explicitly enable the plugin before it's used with the planner. 
- If plugin is enabled, header matching x-sk-copilot-{pluginNameForModel}-auth is added to request for bot response.

New Plugin Gallery look with Import card
image

New card for custom plugin added to gallery
image

Plugin Wizard
image
- If URL is invalid:
image

Validate Manifest Step
image
- If Manifest validation fails:
image

Confirmation
image

Enabling custom Plugin
image

Once user enables plugin, sample request herader:
image

Planner integration
image

### Contribution Checklist

- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md#dev-scripts) raises no violations


webapi/CopilotChat/Controllers/ChatController.cs
```diff
@@ -183,16 +211,47 @@ private async Task RegisterPlannerSkillsAsync(CopilotChatPlanner planner, OpenAp
...
             planner.Kernel.ImportSkill(new TaskListSkill(new MicrosoftToDoConnector(graphServiceClient)), "todo");
             planner.Kernel.ImportSkill(new CalendarSkill(new OutlookCalendarConnector(graphServiceClient)), "calendar");
             planner.Kernel.ImportSkill(new EmailSkill(new OutlookMailConnector(graphServiceClient)), "email");
         }
+
+        if (variables.TryGetValue("customPlugins", out string? customPluginsString))
+        {
+            CustomPlugin[]? customPlugins = JsonSerializer.Deserialize<CustomPlugin[]>(customPluginsString);
+
+            if (customPlugins != null)
```
TaoChenOSU, line 229: nit: debug log for unsuccessful Json deserialization

webapi/CopilotChat/Controllers/ChatController.cs
```diff
@@ -183,16 +211,47 @@ private async Task RegisterPlannerSkillsAsync(CopilotChatPlanner planner, OpenAp
...
+
+                        UriBuilder uriBuilder = new(plugin.ManifestDomain);
+                        // Expected manifest path as defined by OpenAI: https://platform.openai.com/docs/plugins/getting-started/plugin-manifest
+                        uriBuilder.Path = "/.well-known/ai-plugin.json";
+
+                        // TODO: Support other forms of auth. Currently, we only support user PAT or no auth.
+                        var requiresAuth = !plugin.AuthType.Equals("none", StringComparison.OrdinalIgnoreCase);
+                        BearerAuthenticationProvider authenticationProvider = new(() => Task.FromResult(PluginAuthValue));
+
+                        await planner.Kernel.ImportChatGptPluginSkillFromUrlAsync($"{plugin.NameForModel}Plugin", uriBuilder.Uri,
```
TaoChenOSU, line 246: nit:
```suggestion
                        await planner.Kernel.ImportChatGptPluginSkillFromUrlAsync(
                            $"{plugin.NameForModel}Plugin", 
                            uriBuilder.Uri,
```

webapp/src/components/open-api-plugins/PluginGallery.tsx
```diff
@@ -55,40 +58,60 @@ const useClasses = makeStyles({
...
-        <Dialog open={open} onOpenChange={(_event, data) => { setOpen(data.open); }}>
+        <Dialog
+            open={open}
+            onOpenChange={(_event, data) => {
+                setOpen(data.open);
+            }}
+        >
             <DialogTrigger>
-                <Button data-testid="pluginButton" appearance="transparent" icon={<AppsAddIn24Regular color="white" />} />
+                <Button style={{ color: 'white' }} appearance="transparent" icon={<AppsAddIn24 color="white" />}>
```
TaoChenOSU, line 72: `data-testid` is removed. It may be used in scenario tests.

webapi/CopilotChat/Controllers/ChatController.cs
```diff
@@ -137,27 +138,54 @@ public ChatController(ILogger<ChatController> logger, ITelemetryService telemetr
...
     /// Register skills with the planner's kernel.
     /// </summary>
-    private async Task RegisterPlannerSkillsAsync(CopilotChatPlanner planner, OpenApiSkillsAuthHeaders openApiSkillsAuthHeaders, ContextVariables variables)
+    private async Task RegisterPlannerSkillsAsync(CopilotChatPlanner planner, Dictionary<string, string> openApiSkillsAuthHeaders, ContextVariables variables)
     {
         // Register authenticated skills with the planner's kernel only if the request includes an auth header for the skill.
 
         // Klarna Shopping
-        if (openApiSkillsAuthHeaders.KlarnaAuthentication != null)
+        if (openApiSkillsAuthHeaders.TryGetValue("KLARNA", out string? KlarnaAuthHeader))
```
glahaye, line 174: We no longer has constants for KLARNA, GITHUB, JIRA, GRAPH... ?

webapi/CopilotChat/Controllers/ChatController.cs
dehoward: the formatter should have a preference here

webapi/CopilotChat/Models/CustomPlugin.cs
```diff
@@ -0,0 +1,42 @@
...
+/// <summary>
+/// Custom plugin imported from ChatGPT Manifest file.
+/// Docs: https://platform.openai.com/docs/plugins/introduction.
+/// </summary>
+public class CustomPlugin
+{
+    /// <summary>
+    /// Human-readable name, such as the full company name.
+    /// </summary>
+    [JsonPropertyName("nameForHuman")]
```
dehoward, line 15: nit: consider using `displayName` here and just `name` below?

webapp/src/components/open-api-plugins/plugin-wizard/PluginWizard.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+    return (
+        <Dialog
+            onOpenChange={() => {
+                resetLocalState();
+            }}
+            modalType="alert"
+        >
+            <DialogTrigger>
+                <Button data-testid="add-ustom-plugin" aria-label="Add Custom Plugin" appearance="primary">
```
dehoward, line 199: typo

webapp/src/components/open-api-plugins/plugin-wizard/PluginWizard.tsx
```diff
@@ -0,0 +1,227 @@
...
+        setPluginValidated(false);
+    }, []);
+
+    const onAddPlugin = useCallback(() => {
+        if (pluginManifest && manifestDomain) {
+            plugins.addCustomPlugin(pluginManifest, manifestDomain);
+            setActiveStep(CreatePluginSteps.Confirmation);
+        } else {
+            setPluginValidated(false);
+            // TODO: [Issue #1973] add error handling
```
dehoward, line 80: should we update this to an issue in the new repo?

webapp/src/components/utils/PluginUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+            throw new Error(`Invalid api url: ${apiUrl}`);
+        }
+    }
+
+    // If no errors are thrown, the plugin manifest is valid
+    return true;
+};
+
+export const isValidOpenAPISpec = (_specPath: string) => {
+    // TODO: [Issue #1973] Implement validation of OpenAPI spec
```
dehoward, line 62: same here

webapp/src/redux/features/plugins/pluginsSlice.ts
```diff
@@ -1,56 +1,50 @@
...
-                case Plugins.Klarna:
-                    state.Klarna = initialState.Klarna;
-                    break;
+        disconnectPlugin: (state: PluginsState, action: PayloadAction<string>) => {
+            const plugin = state.plugins[action.payload];
+            plugin.enabled = false;
+            plugin.authData = undefined;
+            if (plugin.apiProperties) {
+                Object.keys(plugin.apiProperties).forEach((key) => {
+                    // Falsey lint check, just checked above
```
dehoward, line 35: I was curious as to why and there is a good explanation on it [here](https://stackoverflow.com/questions/57385552/object-is-possibly-undefined-in-typescript). 
the tl/dr is we can (and should) get around this is if we do `const apiProperties = plugin.apiProperties` before the `if`

webapi/CopilotChat/Controllers/ChatController.cs
```diff
@@ -183,16 +211,47 @@ private async Task RegisterPlannerSkillsAsync(CopilotChatPlanner planner, OpenAp
...
+                        UriBuilder uriBuilder = new(plugin.ManifestDomain);
+                        // Expected manifest path as defined by OpenAI: https://platform.openai.com/docs/plugins/getting-started/plugin-manifest
+                        uriBuilder.Path = "/.well-known/ai-plugin.json";
+
+                        // TODO: Support other forms of auth. Currently, we only support user PAT or no auth.
+                        var requiresAuth = !plugin.AuthType.Equals("none", StringComparison.OrdinalIgnoreCase);
+                        BearerAuthenticationProvider authenticationProvider = new(() => Task.FromResult(PluginAuthValue));
+
+                        await planner.Kernel.ImportChatGptPluginSkillFromUrlAsync($"{plugin.NameForModel}Plugin", uriBuilder.Uri,
+                            new OpenApiSkillExecutionParameters
```
craigomatic, line 247: Could we set the `IgnoreNonCompliantErrors` property to true here? 

This will allow importing plugins that our underlying validator does not think are valid (it's overly aggressive unfortunately)

webapi/CopilotChat/Controllers/ChatController.cs
teresaqhoang: These were never constants, only their request headers were. `openApiSkillsAuthHeaders` was a class we used to manually deserialize the right headers (see the change in line 74 of ChatController) from the request, but now we parse all request headers dynamically, so `openApiSkillsAuthHeaders` is now a dictionary object, where the key is extracted from the capturing group in this regex: `x-sk-copilot-(.*)-auth`

This is the only place these values are used in the webap. The only other place they would be needed is in the webapp when the Plugins are defined (see PluginState.ts) but we can't share constants across the two servers, so there's no use in defining them as constants here.

webapp/src/components/open-api-plugins/plugin-wizard/PluginWizard.tsx
teresaqhoang: I don't think it really matters right now since both are active and we're still tracking things on the old board

webapp/src/components/utils/PluginUtils.ts
teresaqhoang: Administrative overhead, will just use the old issue

webapi/CopilotChat/Controllers/ChatController.cs
teresaqhoang: formatted

webapp/src/redux/features/plugins/pluginsSlice.ts
teresaqhoang: Thanks for looking into it! added your suggestion

webapi/CopilotChat/Controllers/ChatController.cs
teresaqhoang: Done! It allowed me import an instacart manifest that failed before. Thanks Craig!! 
image

TaoChenOSU (approved):

dehoward (approved):

alliscode (approved):

alliscode (approved):

PR #31, opened 2023-07-25
📝 RLFH UX Skeleton

teresaqhoang:
### Motivation and Context

Added RLFH UI to demonstrate users can easily add this capability to their chatbot if they wanted.  

### Description

> This feature is for demonstration purposes only. We don't actually hook up to the server to store the human feedback or send it to the model. 

Details:

- RLFH actions will only show on the most recent chat message in which the author is `bot`.
- If user takes action, icon will be rendered to reflect action.
- Data is only stored in frontend state. If app refreshes, all RLFH across all chats will be reset.

Actions on chat message:
image

Once user takes actions:
image

Future work 
- Add ability to turn of RLHF in settings dialog + add help link to docs there

### Contribution Checklist

### DIFF
--- a/webapp/src/components/chat/chat-history/ChatHistoryItem.tsx
+++ b/webapp/src/components/chat/chat-history/ChatHistoryItem.tsx
@@ -2,17 +2,19 @@
 
 import { Persona, Text, makeStyles, mergeClasses, shorthands, tokens } from '@fluentui/react-components';
 import React from 'react';
-import { AuthorRoles, ChatMessageType, IChatMessage } from '../../../libs/models/ChatMessage';
+import { AuthorRoles, ChatMessageType, IChatMessage, UserFeedback } from '../../../libs/models/ChatMessage';
 import { GetResponseOptions, useChat } from '../../../libs/useChat';
 import { useAppSelector } from '../../../redux/app/hooks';
 import { RootState } from '../../../redux/app/store';
-import { Breakpoints } from '../../../styles';
+import { Breakpoints, customTokens } from '../../../styles';
 import { timestampToDateString } from '../../utils/TextUtils';
 import { PlanViewer } from '../plan-viewer/PlanViewer';
 import { PromptDetails } from '../prompt-details/PromptDetails';
 import { ChatHistoryDocumentContent } from './ChatHistoryDocumentContent';
 import { ChatHistoryTextContent } from './ChatHistoryTextContent';
 import * as utils from './../../utils/TextUtils';
+import { ThumbLike16Filled, ThumbDislike24Filled } from '@fluentui/react-icons';
+import { UserFeedbackActions } from './UserFeedbackActions';
 
 const useClasses = makeStyles({
     root: {
@@ -23,6 +25,7 @@ const useClasses = makeStyles({
         ...Breakpoints.small({
             maxWidth: '100%',
         }),
+        ...shorthands.gap(customTokens.spacingHorizontalXS),
     },
     debug: {
         position: 'absolute',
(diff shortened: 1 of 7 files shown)


dehoward:
if we were to send the feedback to the model, do we know if it would have any impact on the responses? without doing this, I'm not sure that this change aligns with our recently updated first principles given that it doesn't seem like we're currently completing the loop on the AI usage pattern.

webapp/src/components/shared/BundledIcons.tsx
```diff
@@ -0,0 +1,46 @@
...
+    Info16Filled,
+    Info16Regular,
+    Share20Filled,
+    Share20Regular,
+    ThumbDislike16Filled,
+    ThumbDislike16Regular,
+    ThumbLike16Filled,
+    ThumbLike16Regular,
+    bundleIcon,
+} from '@fluentui/react-icons';
```
glahaye, line 31: Do we use all of these??

teresaqhoang:
> if we were to send the feedback to the model, do we know if it would have any impact on the responses? without doing this, I'm not sure that this change aligns with our recently updated first principles given that it doesn't seem like we're currently completing the loop on the AI usage pattern.

> if we were to send the feedback to the model, do we know if it would have any impact on the responses? without doing this, I'm not sure that this change aligns with our recently updated first principles given that it doesn't seem like we're currently completing the loop on the AI usage pattern.

No, SK doesn't hook up to any feedback loop with the models, so this would be up to the developer to develop and expose their own endpoints to train their models.

This was a big push from John to showcase of the first principles of LLMs in general `Human in Control`, defined here, https://microsoft.sharepoint.com/teams/ExDAI/SitePages/GPT-UX-Guide.aspx which aligns with our principles to showcase experimental usage patterns in Microsoft.

webapp/src/components/shared/BundledIcons.tsx
teresaqhoang: Yeah these are all currently used, so easiest to port them all over now

dehoward:
> No, SK doesn't hook up to any feedback loop with the models, so this would be up to the developer to develop and expose their own endpoints to train their models.
> 
> This was a big push from John to showcase of the first principles of LLMs in general `Human in Control`, defined here, https://microsoft.sharepoint.com/teams/ExDAI/SitePages/GPT-UX-Guide.aspx which aligns with our principles to showcase experimental usage patterns in Microsoft. Even though we don't complete the loop, it's beneficial to call out this principal of augmentation over automation. Planning to add a help link in later PR with settings that will detail this process more.

ok yeah that makes sense. thanks for sharing this, video is very insightful. 

> Planning to add a help link in later PR with settings that will detail this process more.

I think this will help a lot, especially if we can emphasize some of what you said here in that they will need to train their own models.

webapp/src/components/chat/chat-history/ChatHistoryItem.tsx
```diff
@@ -91,6 +94,13 @@ export const ChatHistoryItem: React.FC<ChatHistoryItemProps> = ({ message, getRe
         content = <ChatHistoryTextContent message={message} />;
     }
 
+    // TODO: hookup to backend
```
dehoward, line 97: when we leave TODOs in the app, are they:
1) for our team to follow-up on?
2) for teams taking CC as their own to later follow-up on?
3) both?

dehoward (approved):

TaoChenOSU (approved):

webapp/src/components/chat/chat-history/ChatHistoryItem.tsx
teresaqhoang: No official guidance but our team seems to follow both. In general, if you add a TODO, you're expected to follow-up on it or change it to a comment describing the gap in development / expected behavior. Some have slipped through the cracks with the dev velocity on CC so we should probably clean those up with this migration

dehoward: > Some have slipped through the cracks with the dev velocity on CC so we should probably clean those up with this migration

I agree that a clean-up is likely in order. and with that it would be nice to have a clear distinction between TODOs left for us to follow-up on (which should usually include a Task #) vs those left for an inheriting/customer dev to take on... I can create a backlog item for this.

craigomatic:
> > No, SK doesn't hook up to any feedback loop with the models, so this would be up to the developer to develop and expose their own endpoints to train their models.

Is there a simple implementation we can being in? 

If we have buttons on the UI that are not wired up we are going to end up with people wondering why they aren't working.

Simple could be using them as few shot examples in conjunction with the chat message history? Could add a state to each message?

PR #34, opened 2023-07-25
Add documents tab

TaoChenOSU:
### Motivation and Context

Refactor files tab to documents tab.

### Description
1. Enable document import in the documents tab.
2. Show progress on each document importing.
3. Show token usage of each imported document.
4. Move alerts to the bottom of the screen.
5. 5. Add support for markdown files.

image
image
image
image

### Contribution Checklist

### DIFF
--- a/webapi/CopilotChat/Controllers/DocumentImportController.cs
+++ b/webapi/CopilotChat/Controllers/DocumentImportController.cs
@@ -49,6 +49,11 @@ public class DocumentImportController : ControllerBase
         Pdf,
 
         /// <summary>
+        /// .md
+        /// </summary>
+        Md,
+
+        /// <summary>
         /// .jpg
         /// </summary>
         Jpg,
@@ -254,15 +259,7 @@ public class DocumentImportController : ControllerBase
             }
 
             // Make sure the file type is supported.
-            var fileType = this.GetFileType(Path.GetFileName(formFile.FileName));
-            switch (fileType)
-            {
-                case SupportedFileType.Txt:
-                case SupportedFileType.Pdf:
-                    break;
-                default:
-                    throw new ArgumentException($"Unsupported file type: {fileType}");
-            }
+            _ = this.GetFileType(Path.GetFileName(formFile.FileName));
         }
     }
 
@@ -280,6 +277,7 @@ public class DocumentImportController : ControllerBase
         switch (fileType)
         {
             case SupportedFileType.Txt:
+            case SupportedFileType.Md:
                 documentContent = await this.ReadTxtFileAsync(formFile);
                 break;
             case SupportedFileType.Pdf:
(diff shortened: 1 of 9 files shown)


webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -248,15 +259,7 @@ private async Task ValidateDocumentImportFormAsync(DocumentImportForm documentIm
...
-            var fileType = this.GetFileType(Path.GetFileName(formFile.FileName));
-            switch (fileType)
-            {
-                case SupportedFileType.Txt:
-                case SupportedFileType.Pdf:
-                    break;
-                default:
-                    throw new ArgumentException($"Unsupported file type: {fileType}");
-            }
+            _ = this.GetFileType(Path.GetFileName(formFile.FileName));
```
glahaye, line 262: Nit: Might want to put a comment here to say this is done to verify file type is supported and that, otherwise, an exception is raise

dehoward, line 262: nit: the assignment can be removed here

webapp/src/components/chat/ChatResourceList.tsx
```diff
@@ -56,24 +79,117 @@ interface TableItem {
...
+            : [];
+
+        setDocumentImporting(importingResources.length > 0);
+        setResources(importingResources);
+
+        void chat.getChatMemorySources(selectedId).then((sources) => {
+            setResources([...importingResources, ...sources]);
         });
+
         // eslint-disable-next-line react-hooks/exhaustive-deps
```
dehoward, line 118: looks like we can remove this?

webapp/src/components/chat/ChatResourceList.tsx
```diff
@@ -56,24 +79,117 @@ interface TableItem {
...
+                    <Button
+                        className={classes.uploadButton}
+                        icon={<DocumentArrowUp20Regular />}
+                        disabled={documentImporting}
+                        onClick={() => documentFileRef.current?.click()}
+                    >
+                        Upload
+                    </Button>
+                </Tooltip>
+                {documentImporting && <Spinner size="tiny" />}
```
dehoward, line 177: since `importingDocuments` an `documentImporting` are always set together, it seems like we could just use `importDocuments` in place of `documentImporting`?

```suggestion
                {importDocuments.length && <Spinner size="tiny" />}
```

webapp/src/components/chat/ChatResourceList.tsx
```diff
@@ -56,24 +79,117 @@ interface TableItem {
...
+                {/* Hack: Hardcode vector database. */}
```
dehoward, line 178: should we give some indication in the UI as to why these options are hardcoded?

webapp/src/components/chat/ChatWindow.tsx
```diff
@@ -249,7 +214,10 @@ export const ChatWindow: React.FC = () => {
                 </div>
             </div>
             {selectedTab === 'chat' && <ChatRoom />}
-            {selectedTab === 'files' && <ChatResourceList chatId={selectedId} />}
+            {selectedTab === 'documents' && <ChatResourceList />}
+            {selectedTab !== 'chat' && <div className={classes.alerts}>
+                <Alerts />
+            </div>}
```
dehoward, line 220: we're adding other tabs soon so this will apply to multiple pages, right?

TaoChenOSU: Yes, except the Chat tab where the alert will be above the input box.

webapp/src/components/chat/ChatResourceList.tsx
TaoChenOSU: We still need this as we don't want to depend on `chat`

dehoward (approved):

PR #39, opened 2023-07-26
Token Usage

teresaqhoang:
### Motivation and Context

This PR adds the token usage feature, in which token usage is calculated and shown per prompt and per session. Each token usage calculation will be split into two values:
1. total tokens used in chat completion of the bot response prompt
2. total tokens used in dependencies used to generate prompt

### Description

webapi
- Token usage per prompt persists as part of ChatMessage object
- Add initial bot message and token usage tracking to ChatHistoryController and tracking of token usage for bot response in ChatMessage and ChatSession models.
- Update ChatSkill to save token usage to context variables and return with bot response. 
- Added token usage calculation in ChatSkill by calculating total token usage for dependency functions and chat completion and sending updated response to client. Copy token usage into original chat context. 
- Calculate memory extraction token usage by taking into account cumulative semantic memory token usage. 
- Update Utilities to include GetTokenUsage method.

Webapp
- AppState: Added a new TokenUsage field to track total usage across all chats by app session, and appSlice has been updated to cumulate session token usage.
- Update SignalRMiddleware to handle token usage when receiving message updates from server. Update message property to tokenUsage if tokenUsage is defined, otherwise update content. 
- Fix ChatHistoryTextContent to include TypingIndicator when bot response is generating. 
- Changed PromptDetails -> PromptDialog component to show prompt details and token usage graph. 
- Removed TypingIndicatorRenderer

Token usage shown for ChatMessages of type Message and Plan
image

Hardcoded bot responses default to 0
image
Loading state
image

Info
image

This is what it will look like in settings dialog once changes go in
image

### Contribution Checklist

~~[ ] All unit tests pass, and I have added new tests where possible~~


webapi/CopilotChat/Models/ChatMessage.cs
```diff
@@ -110,6 +110,12 @@ public enum ChatMessageType
     [JsonPropertyName("type")]
     public ChatMessageType Type { get; set; }
 
+    /// <summary>
+    /// Counts of total token usage used to generate bot response.
+    /// </summary>
+    [JsonPropertyName("tokenUsage")]
+    public TokenUsage? TokenUsage { get; set; }
```
dehoward, line 117: if we use a default for this then we shouldn't need to pass it in other places
```suggestion
    public TokenUsage? TokenUsage { get; set; } = new TokenUsage(0, 0);
```

webapi/CopilotChat/Models/ChatSession.cs
```diff
@@ -29,6 +29,12 @@ public class ChatSession : IStorageEntity
     [JsonPropertyName("createdOn")]
     public DateTimeOffset CreatedOn { get; set; }
 
+    /// <summary>
+    /// Initial bot message.
+    /// </summary>
+    [JsonPropertyName("initialBotMessage")]
+    public ChatMessage? InitialBotMessage { get; set; }
```
TaoChenOSU, line 36: What is the purpose of saving the initial message in the session? This couples the ChatMessage and ChatSession models.

webapi/CopilotChat/Skills/Utilities.cs
```diff
@@ -29,4 +31,9 @@ internal static SKContext CopyContextWithVariablesClone(SKContext context)
     /// Calculate the number of tokens in a string.
     /// </summary>
     internal static int TokenCount(string text) => GPT3Tokenizer.Encode(text).Count;
+
+    /// <summary>
+    /// Get the total toal usage from ChatCompletion result context.
```
dehoward, line 36: typo

webapi/CopilotChat/Models/ChatMessage.cs
TaoChenOSU, line 117: I think life will be probably easier if token usage cannot be null.

webapi/CopilotChat/Controllers/ChatHistoryController.cs
```diff
@@ -80,13 +80,14 @@ public async Task<IActionResult> CreateChatSessionAsync([FromBody] CreateChatPar
...
+        // Create initial bot message
         var chatMessage = ChatMessage.CreateBotResponseMessage(
             newChat.Id,
-            initialBotMessage,
-            string.Empty);
+            this._promptOptions.InitialBotMessage,
+            string.Empty, // The initial bot message doesn't need a prompt.
+            new TokenUsage(0, 0));
         await this._messageRepository.CreateAsync(chatMessage);
+        newChat.InitialBotMessage = chatMessage;
```
TaoChenOSU, line 90: Why are we duplicating the message in storage?

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -137,6 +137,8 @@ public async Task<string> ExtractUserIntentAsync(SKContext context)
             settings: this.CreateIntentCompletionSettings()
         );
 
+        context.Variables.Set("userIntentExtractionTokenUsage", Utilities.GetTokenUsage(result).ToString(CultureInfo.InvariantCulture));
```
TaoChenOSU, line 140: nit: use an enum to define the variable names.

webapp/src/redux/features/app/AppState.ts
```diff
@@ -17,3 +18,31 @@ export interface Alert {
...
+}
+
+export interface AppState {
+    alerts: Alert[];
+    activeUserInfo?: ActiveUserInfo;
+    // Total usage across all chats by app session
+    tokenUsage: TokenUsage;
+}
+
+export const initialState: AppState = {
```
dehoward, line 36: nit: `InitialAppState`

webapi/CopilotChat/Skills/Utilities.cs
```diff
@@ -29,4 +31,9 @@ internal static SKContext CopyContextWithVariablesClone(SKContext context)
...
+    /// </summary>
+    internal static int GetTokenUsage(SKContext result) => result.ModelResults.First().GetResult<ChatCompletions>().Usage.TotalTokens;
```
TaoChenOSU, line 38: nit: validate if ModelResults contains any element before calling `First()`

webapp/src/components/shared/TokenUsageGraph.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    const hasTokenCount = (promptUsage ?? dependencyUsage) !== undefined;
+    const loadingResponse = conversations[selectedId].botResponseStatus;
+
+    // Necessary conversion due to type coercion issues
+    promptUsage = Number(promptUsage);
+    dependencyUsage = Number(dependencyUsage);
+
+    const classes = useClasses();
+    const maxWidth = 500;
```
TaoChenOSU, line 57: nit: use a constant

webapp/src/libs/hooks/useChat.ts
```diff
@@ -123,12 +122,28 @@ export const useChat = () => {
...
+                    ask,
+                    await AuthHelper.getSKaaSAccessToken(instance, inProgress),
+                    getEnabledPlugins(),
+                )
+                .catch((e: any) => {
+                    throw e;
+                });
+
+            // Update token usage of current session
+            const promptTokenUsage = askResult.variables.find((v) => v.key === 'promptTokenUsage')?.value;
```
TaoChenOSU, line 136: nit: define the names in an enum

webapi/CopilotChat/Models/ChatMessage.cs
teresaqhoang: This was intentional to force token usage calculation to be explicit when creating new bot response message. If it hasn't been explicitly calculated, we show an unknown state since 0 isn't truthy for all scenarios.

webapp/src/redux/features/app/AppState.ts
teresaqhoang: I believe initialState is the standard practice, so we can just pass the variable as is to createSlice rather than explicitly naming in every feature slice like this

image

webapi/CopilotChat/Models/ChatSession.cs
teresaqhoang: It helps optimize our network calls from the webapp (why load the messages when we already know what it is)? What's the reservation with coupling the two models?

webapi/CopilotChat/Models/ChatMessage.cs
dehoward: oh I see, since some responses are hardcoded and are supposed to have 0 token usage vs. sometimes we may fail to calculate it?

webapi/CopilotChat/Models/ChatSession.cs
teresaqhoang: Optionally, I can create a new model called CreateChatResponse that's not coupled to a storage repository

webapi/CopilotChat/Models/ChatMessage.cs
teresaqhoang: This was intentional to force token usage calculation to be explicit when creating new bot response message. If it hasn't been explicitly calculated, we show an unknown state since 0 isn't truthy for all scenarios. -- sets a better standard for token management

teresaqhoang: Desmond left a similar comment up top, so let's continue that convo up there. Resolving this convo as a dupe

webapi/CopilotChat/Controllers/ChatHistoryController.cs
teresaqhoang: It's not stored -- the CreateAsync method is called before

webapp/src/components/shared/TokenUsageGraph.tsx
teresaqhoang: It is a constant, it's a style that only applies to this component so not need to add it at a larger scale

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: C# doesn't easily support string enums so I used a dict

webapp/src/redux/features/app/AppState.ts
dehoward: that's fair, but if needed elsewhere it can make for confusing imports:
```ts
import { initialState } from './AppState';
import { initialState } from './ConverstationsState';
```

I don't have a strong opinion on this, although I'd say we should generally lean toward clear variable naming vs. convenience in assignment.

teresaqhoang: We don't use initial state anywhere else, so it's well defined in the context since we follow a well-structured file naming pattern for redux features. Will leave as is

webapi/CopilotChat/Models/ChatSession.cs
TaoChenOSU: I think it's better to not save an entry of the ChatMessage model in the ChatSession model. This will have three disadvantages with the only benefit of reducing one single network call:
1. It essentially creates two copies of the initial message in the database, one in the session repository, the other in the message repository.
2. It breaks the pattern of loading/handling sessions and messages separately. 
3. It limits our ability to create multiple initial messages (not likely to happen but imagine if we want an introduction message then a greeting message).

webapi/CopilotChat/Controllers/ChatHistoryController.cs
TaoChenOSU: You are saving the message in the message repository as well as the session repository.

webapp/src/libs/hooks/useChat.ts
```diff
@@ -123,12 +123,32 @@ export const useChat = () => {
...
+                .catch((e: any) => {
+                    throw e;
+                });
+
+            // Update token usage of current session
+            const promptTokenUsage = askResult.variables.find(
+                (v) => v.key === (TokenUsageKeys.prompt as string),
+            )?.value;
+            const dependencyTokenUsage = askResult.variables.find(
+                (v) => v.key === (TokenUsageKeys.prompt as string),
```
dehoward, line 141:
```suggestion
                (v) => v.key === (TokenUsageKeys.dependency as string),
```

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -660,6 +695,35 @@ private int GetChatContextTokenLimit(string audience, string userIntent)
...
+    /// <param name="response">Message result of chat completion</param>
+    /// <param name="chatContext">Context maintained during response generation.</param>
+    private TokenUsage CalculateTokenUsage(string chatId, string userId, string response, SKContext chatContext)
+    {
+        var prompt = chatContext.Variables.TryGetValue("prompt", out string? renderedPrompt)
+            ? renderedPrompt
+            : string.Empty;
+
+        // Token usage of chat completion = prompt + completion
+        var chatCompletionTokenUsage = Utilities.TokenCount(prompt) + Utilities.TokenCount(response);
```
dehoward, line 712: this doesn't seem right. we're showing this as the token usage of the "Prompt" in the UX, but here we include the tokens of the completion as well. look at OAI's interface for [CompletionsUsage](https://learn.microsoft.com/en-us/dotnet/api/azure.ai.openai.completionsusage?view=azure-dotnet-preview), I would expect that we show "Prompt" and "Completion" token usage separately?

teresaqhoang: The token usage shown in the UI is the total token usage for the prompt completion. Essentially with any completion request, total token usage = tokens in the prompt + tokens in the response. In the image below, the first "Hi! Can you help me plan a trip to Hawaii?" is considered the prompt (call) and the green part is the response (completion).
image

Since the chat streaming changes weren't backwards compatible with the token usage implementation in the feature branch, there isn't an easy way to get the total token usage from the function result as we can when just calling the chatCompletion function, so I had to calculate it manually like this. 

I previously aligned with John that we should just show the total usage (rather than splitting it up into prompt and completion for dependencies and bot response because this can get tricky to communicate well in the UI. 

I changed the copy in the info bubble to define this better
image

webapi/CopilotChat/Models/TokenUsage.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        if (result.ModelResults == null || result.ModelResults.Count == 0)
+        {
+            chatContext.Log.LogError("Unable to determine token usage for {0}", functionKey);
+            return;
+        }
+
+        int previousSum = includePrevious && chatContext.Variables.TryGetValue(functionKey!, out string? cumulativeTokenUsage) ? int.Parse(cumulativeTokenUsage, CultureInfo.CurrentCulture) : 0;
+
+        var tokenUsage = result.ModelResults.First().GetResult<ChatCompletions>().Usage.TotalTokens;
```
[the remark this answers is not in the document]
teresaqhoang: They're meant to be double counted since the LLM processes the dependency completion result again when tokenizing the final prompt. 

i.e., in UserIntent, if the total tokens are 450 and 350 was from the userIntentPrompt and 100 was from the completion, that means the LLM used 450 to process and complete the prompt. 

And then since we only use the user intent response (completion) in the final bot response prompt, we need to account for that in the "prompt" token count of the final prompt because the LLM processes that again.

webapi/CopilotChat/Controllers/ChatHistoryController.cs
teresaqhoang: Removed initialBotMessage from ChatSession object

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
dehoward: > I previously aligned with John that we should just show the total usage (rather than splitting it up into prompt and completion for dependencies and bot response because this can get tricky to communicate well in the UI.

what happened after this? I'm guessing John is on board with these changes?

> I changed the copy in the info bubble to define this better

for `"Prompt token usage"`, I think that `"the total number of tokens used in the completion of the bot generation prompt"` is correct, but we also should mention that it includes the number of tokens from the actual prompt as well, since right now we only mention the "completion of" the prompt. 

I'm starting to see John's point that this can be tricky to show 😄.

really tho, I think we should be more explicit in the naming here to not confuse users with OAI's definition of these terms. which for prompt token usage, [only includes](https://learn.microsoft.com/en-us/dotnet/api/azure.ai.openai.completionsusage.prompttokens?view=azure-dotnet-preview#azure-ai-openai-completionsusage-prompttokens) the `"tokens in the provided prompts for the completion request"`, and not those in the completion itself. 

I'm thinking we make the following changes, let me know what you think:
1) show `"Prompt generation token usage"` as the `"total number of tokens used in the generation of the prompt and its final output"` (which is the sum of `TotalTokens` from the dependency functions, and the token count from the prompt).
2) show "Completion token usage" as the `"total number of tokens used in the completion of the response"`. 
3 and potentially show "Total token usage" as the sum of these two.

webapi/CopilotChat/Models/TokenUsage.cs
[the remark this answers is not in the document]
dehoward: that makes sense, thanks.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: Consulted with expert @matthewbolanos and we decided to get rid of the "Dependencies" bucket altogether and instead, be more specific with the token usage and split it up by explicit completions:

1. Extract audience tokens
1. Extract user intent tokens
1. Bot response tokens
1. Generate Semantic Memory tokens

Each of these counts will be total token usage (prompt + completion), so this will help remove the ambiguity of an overloaded "prompt" term

webapi/CopilotChat/Models/ChatMessage.cs
TaoChenOSU: Could we then have an enum in the TokenUsage class that validates the state of the token usage? i.e. `Calculated, Hardcoded, Unknown`

TaoChenOSU: This is not a requirement but for better code readability. Allowing something to be null may need additional code to handle checks and whatnot. And it sounds like null actually means something in this case (unknown token usage), we can better convey that idea by setting an unknown state. this can be done in a separate PR if this PR needs to be merged ASAP.

TaoChenOSU (approved):

webapi/Services/IOcrEngine.cs
```diff
@@ -10,7 +10,6 @@ namespace SemanticKernel.Service.Services;
 /// </summary>
 public interface IOcrEngine
 {
-
```
crickman, line 13: Intentional?  (Perhaps this file can be reverted)

webapi/CopilotChat/Skills/TokenUtilities.cs
```diff
@@ -0,0 +1,94 @@
...
+        { "SystemCompletion", "responseCompletion"}
+    };
+
+    /// <summary>
+    /// Gets dictorary containing empty token usage totals.
+    /// Use for responses that are hardcoded and/or do not have semantic (token) dependencies.
+    /// </summary>
+    internal static Dictionary<string, int> EmptyTokenUsages()
+    {
+        var emptyTokenUsages = new Dictionary<string, int>();
```
crickman, line 39: semanticFunctions.Values.ToDictionary(v => v, v => 0);

Also please verify whether you intend to dictionary key comparison be case invariant or not. (currently not).

crickman (dismissed):

webapp/src/components/chat/chat-history/ChatHistoryTextContent.tsx
```diff
@@ -19,14 +20,18 @@ interface ChatHistoryTextContentProps {
...
-            return `&#${i.charCodeAt(0)};`;
-        });
+    let content = message.content.trim().replace(/[\u00A0-\u9999<>&]/g, function (i: string) {
+        return `&#${i.charCodeAt(0)};`;
+    });
     content = utils.formatChatTextContent(content);
-    content = content
-            .replace(/\n/g, '<br />')
-            .replace(/ {2}/g, '&nbsp;&nbsp;');
+    content = content.replace(/\n/g, '<br />').replace(/ {2}/g, '&nbsp;&nbsp;');
```
TaoChenOSU, line 27: nit: comments on what these two statements do

webapp/src/components/token-usage/TokenUsageGraph.tsx
```diff
@@ -0,0 +1,159 @@
...
+    Object.entries(tokenUsage).forEach(([key, value]) => {
+        const viewDetails: TokenUsageViewDetails = {
+            usageCount: value ?? 0,
+            legendLabel: TokenUsageFunctionNameMap[key],
+            dependency: false,
+            color: semanticKernelBrandRamp[color],
+        };
+
+        // Iterate through the color ramp
+        color -= 20;
```
TaoChenOSU, line 75: nit: prevent it to go out of bound
```suggestion
        color = (color - 20 < 0) ? 160 : color - 20;
```

TaoChenOSU (dismissed):

webapi/CopilotChat/Models/CreateChatResponse.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace SemanticKernel.Service.CopilotChat.Models;
+
+/// <summary>
+/// Response to chatSession/create request.
+/// </summary>
+public class CreateChatResponse
+{
+    /// <summary>
+    /// Chat ID that is persistent and unique to.
```
dehoward, line 13: incomplete sentence?

webapi/CopilotChat/Skills/TokenUtilities.cs
```diff
@@ -0,0 +1,94 @@
...
+            emptyTokenUsages.Add(key, 0);
+        }
+
+        return emptyTokenUsages;
+    }
+
+    /// <summary>
+    /// Gets key used to identify function token usage in context variables.
+    /// </summary>
+    /// <param name="chatContext">Context maintained during response generation.</param>
```
dehoward, line 52: update XML comment

webapp/src/components/token-usage/TokenUsageBar.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface ITokenUsageBar {
+    functionName: string;
+    functionUsage: number;
+    totalUsage: number;
+    color: string;
+}
+
+export const TokenUsageBar: React.FC<ITokenUsageBar> = ({ functionName, functionUsage, totalUsage, color }) => {
+    const percentage = functionUsage / totalUsage;
+    const barWidth = percentage * Constants.MAX_BAR_GRAPH_WIDTH;
```
dehoward, line 13: I don't think we need an app-level constant for this, seems more appropriate as a file-level constant. especially considreing that the max width will depend on how/where the bar is rendered?

dehoward (changes requested):

webapi/Services/IOcrEngine.cs
teresaqhoang: I didn't edit this file, so this was a dotnet format autofix -- will leave as is

webapp/src/components/chat/chat-history/ChatHistoryTextContent.tsx
teresaqhoang: This was a format autofix so not in the scope of my PR

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -405,9 +416,13 @@ public async Task<string> ExtractAudienceAsync(SKContext context)
...
         }
 
+        // Calculate total token usage for dependency functions and prompt template and send with chat stream to client
+        await this.UpdateBotResponseStatusOnClient(chatId, "Calculating token usage");
+        chatContext.Variables.Set(TokenUtilities.GetFunctionKey(chatContext.Log, "SystemMetaPrompt")!, TokenUtilities.TokenCount(renderedPrompt).ToString(CultureInfo.InvariantCulture));
+
         // Stream the response to the client
         await this.UpdateBotResponseStatusOnClient(chatId, "Generating bot response");
-        var chatMessage = await this.StreamResponseToClient(chatId, userId, renderedPrompt);
+        var chatMessage = await this.StreamResponseToClient(chatId, userId, renderedPrompt, await this.GetTokenUsagesAsync(chatContext));
```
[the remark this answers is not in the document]
teresaqhoang: It doesn't change the Big O work or time complexity so it doesn't really matter either way but I removed the initial update

webapp/src/components/token-usage/TokenUsageBar.tsx
teresaqhoang: Good catch, I was sharing it with another component but reverted the changes. Removed

webapi/CopilotChat/Skills/TokenUtilities.cs
teresaqhoang: oh awesome, didn't know about this syntax `.Values.ToDictionary(v => v, v => 0);`. Thanks! Added comparison definition too.

webapi/CopilotChat/Models/ChatMessage.cs
teresaqhoang: Unknown state is no longer relevant since we the entire ChatAsync method fails if any dependencies fail, so it's not shown in chat history so we don't need to account for it. 

So there's only 2 states, Calculated and Hardcoded, which is already handled with the EmptyTokenUsages() definition.
image

TaoChenOSU (approved):

PR #52, opened 2023-07-26
Add Labeler workflow

dehoward:
adds workflow to enable the Labeler action


dehoward:
this PR is dependent on an owner (@alliscode) adding the secret reference in the workflow

alliscode (approved):

PR #65, opened 2023-07-28
Update CC to SK 0.18

glahaye:
### Motivation and Context
The world evolves and so must we.

### Description
Adapted CC to SK 0.18:
- SKContext no longer contains Memory nor a cancellation token, so now using Memory from Kernel
- Make use of SKContext.Clone() and remove a redundant implementation
- Removed code related to ACS semantic search, which no longer exists (only ACS vector search remains)

### Contribution Checklist


crickman (dismissed):
Nice how this is an overall simplification.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -424,21 +424,13 @@ public async Task<string> ExtractAudienceAsync(SKContext context)
...
-
-        var audienceContext = new SKContext(
-            contextVariables,
-            context.Memory,
-            context.Skills,
-            context.Log,
-            context.CancellationToken
-        );
+        SKContext audienceContext = context.Clone();
+        audienceContext.Variables.Set("chatId", context["chatId"]);
```
TaoChenOSU, line 433: If we are cloning the entire context, do we still need to set the variables? Same goes to other methods.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -484,7 +468,7 @@ private async Task<string> GetUserIntentAsync(SKContext context)
     }
 
     /// <summary>
-    /// Helper function create the correct context variables to
+    /// Helper function that creates the correct context variables to
```
TaoChenOSU, line 471: This function doesn't create a new context anymore.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -493,11 +477,11 @@ private async Task<string> GetUserIntentAsync(SKContext context)
     /// <param name="tokenLimit">Maximum number of tokens.</param>
     private Task<string> QueryChatMemoriesAsync(SKContext context, string userIntent, int tokenLimit)
     {
-        return this._semanticChatMemorySkill.QueryMemoriesAsync(userIntent, context["chatId"], tokenLimit, context.Memory);
+        return this._semanticChatMemorySkill.QueryMemoriesAsync(userIntent, context["chatId"], tokenLimit, this._kernel.Memory);
     }
 
     /// <summary>
-    /// Helper function create the correct context variables to
+    /// Helper function that creates the correct context variables to
```
TaoChenOSU, line 484: This also doesn't create a new context.

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
```diff
@@ -88,12 +88,7 @@ public class ExternalInformationSkill
             string planJson = JsonSerializer.Serialize(deserializedPlan.Plan);
             // Reload the plan with the planner's kernel so
             // it has full context to be executed
-            var newPlanContext = new SKContext(
-                null,
-                this._planner.Kernel.Memory,
-                this._planner.Kernel.Skills,
-                this._planner.Kernel.Log
-            );
+            var newPlanContext = new SKContext(null, this._planner.Kernel.Skills, this._planner.Kernel.Log);
```
TaoChenOSU, line 91: Does the planner require memory? If so, does the kernel for the planner point to the same memory store as the other kernel? @teresaqhoang

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
glahaye: Thanks for this.  Fixed.

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
glahaye: Fixed

webapi/CopilotChat/Skills/ChatSkills/ChatSkill.cs
glahaye: fixed

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
glahaye: If it does, it must be using the memory from kernel.Memory...

webapi/SemanticKernelExtensions.cs
```diff
@@ -143,34 +141,19 @@ private static void AddSemanticTextMemory(this IServiceCollection services)
...
-                }
-                // ACS's semantic search where ACS calculates the embeddings
-                else
+                services.AddSingleton<IMemoryStore>(sp =>
                 {
-                    services.AddSingleton<ISemanticTextMemory>(sp => new AzureCognitiveSearchMemory(config.AzureCognitiveSearch.Endpoint, config.AzureCognitiveSearch.Key));
-                }
+                    return new AzureCognitiveSearchMemoryStore(config.AzureCognitiveSearch.Endpoint, config.AzureCognitiveSearch.Key);
+                });
+                services.AddScoped<ISemanticTextMemory>(sp => new SemanticTextMemory(
```
TaoChenOSU, line 148: nit: move this out to reduce duplication.

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
TaoChenOSU: Yeah, but the planner uses a kernel that is not the one used by chat.

glahaye: If it's obtained through dependency injection, it should be the same singleton instance

webapi/SemanticKernelExtensions.cs
glahaye: Done

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
glahaye: So I just double-checked. The ActionPlanner doesn't use memory. The sequential one does to generate and store embeddings for the function descriptions.

Also, you were right, the planner kernel wasn't assigned the memory from dependency injection.

I'm making the necessary changes so that we are goo when dynamically adding functions to the kernel.

teresaqhoang: The Planner doesn't use memory, but we're planning to default to SequentialPlanner, and there's talk of using the planner to load memory as needed, as opposed to in each chat request, so I would recommend at least putting a TODO in here to pass memory into planner. Check with Ben on what issue would be best to tag.

glahaye: OK, then.  I'll just give the planner access to memory through dependency injection (one line changed) but won't change the current planner behavior. If and when we do, memory will be there.

alliscode (approved):

PR #80, opened 2023-08-01
[Planner Stabilization] Adding retry logic + renaming custom Skills -> Plugins

teresaqhoang:
### Motivation and Context

This pull request refactors the GitHub and Jira Skills to Plugins, updates PlannerOptions to include a RelevancyThreshold option, updates ExternalInformationSkill to include retry logic for invalid plans, and improves the memory generation usage in the TokenUsageGraph component.

### Description

- Refactored the GitHubSkill and JiraSkill model classes to GitHubPlugin and JiraPlugin, respectively.
- Added RelevancyThreshold option to PlannerOptions, which is a double value between 0 and 1.0, to control configuration of SequentialPlanner.
- Added retry logic for invalid plans to ExternalInformationSkill, which allows the plan to be retried if the LLM returns an invalid plan.
- Renamed memory generation -> memory extraction.

### Contribution Checklist


dehoward:
let's go ahead and rename `ChatSkills/` to `ChatPlugins/` and also do the same for each "skill" in the folder? I can't imagine that we want both terms floating around.

webapi/CopilotChat/Options/PlannerOptions.cs
```diff
@@ -16,4 +16,16 @@ public class PlannerOptions
...
     public PlanType Type { get; set; } = PlanType.Action;
+
+    /// <summary>
+    /// Whether to retry plan creation if LLM returned response that doesn't contain valid plan (invalid XML or JSON).
+    /// </summary>
+    public bool AllowRetriesOnInvalidPlans { get; set; } = false;
+
+    /// <summary>
+    /// The minimum relevancy score for a function to be considered during plan creation
+    /// when using SequentialPlanner
```
dehoward, line 27: super nit: add a period here and put on line 26?

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
```diff
@@ -118,7 +118,25 @@ public class ExternalInformationSkill
...
+                    plan = await this._planner.CreatePlanAsync($"Given the following context, accomplish the user intent.\nContext:\n{contextString}\nUser Intent:{userIntent}");
+                }
+                catch (PlanningException e) when (e.ErrorCode == PlanningException.ErrorCodes.InvalidPlan && this._planner.PlannerOptions!.AllowRetriesOnInvalidPlans)
+                {
+                    if (maxRetries-- > 0)
+                    {
+                        // Retry plan creation if LLM returned response that doesn't contain valid plan (invalid XML or JSON).
+                        context.Log.LogWarning("Retrying CreatePlan on error: {0}", e.Message);
+                        continue;
+                    }
```
dehoward, line 137: it seems like we need a `break` after this conditional to prevent an infinite loop?

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
```diff
@@ -118,7 +118,25 @@ public class ExternalInformationSkill
         {
             // Create a plan and set it in context for approval.
             var contextString = string.Join("\n", context.Variables.Where(v => v.Key != "userIntent").Select(v => $"{v.Key}: {v.Value}"));
-            Plan plan = await this._planner.CreatePlanAsync($"Given the following context, accomplish the user intent.\nContext:\n{contextString}\nUser Intent:{userIntent}");
+            Plan? plan = null;
+            int maxRetries = 1;
```
dehoward, line 122: should this be controllable by the user? having the max set as 1 seems to defeat the purpose of the loop below.

dehoward (changes requested):
yay for stablization 🥳

webapi/CopilotChat/Skills/OpenApiPlugins/README.md
```diff
@@ -1,26 +1,29 @@
...
-This description is extracted from GitHub's official OpenAPI descriptions for their REST APIs, which can be found in 
+## GitHubPlugin
+
+The OpenAPI spec at `./GitHubPlugin/openapi.json` defines the APIs for GitHub operations
+[List Pull Requests](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests) and
+[Get Pull Request](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request).
+This description is extracted from GitHub's official OpenAPI descriptions for their REST APIs, which can be found in
 [this repository](https://github.com/github/rest-api-description/blob/main/descriptions/ghec/ghec.2022-11-28.json).
 
 ## KlarnaSkill
```
glahaye, line 11:
```suggestion
## KlarnaPlugin
```

webapi/CopilotChat/Skills/OpenApiPlugins/README.md
```diff
@@ -1,26 +1,29 @@
...
+
 The OpenAPI spec at `./KlarnaSkill/openapi.json` defines the APIs for Klarna Shopping's ChatGPT AI plugin operations.
```
glahaye, line 13: Humm... The path is still KlarnaSkill???

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
teresaqhoang: I don't think we should ever run into that case because any exception or error in the block will force the CreatePlanAsync method to bail out, but I added a break as an extra safeguard

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
teresaqhoang: This should not be user-configurable since this retry is a temporary stabilization until the root causes are fixed in the Planner, so we need to heed any retry logic with warning. Aligned with Nilesh that we should limit this kind of retry to 1 so we can accurately identify SK failures.

teresaqhoang:
> let's go ahead and rename `ChatSkills/` to `ChatPlugins/` and also do the same for each "skill" in the folder? I can't imagine that we want both terms floating around.

@dehoward In general, we should keep PRs compartmentalized to their purpose, so this is out of scope here. The change included in this PR was to fix a regression introduced with CustomPlugins that broke the GitHub and Jira plugins when called as plan steps.

There's an effort team-wide to rename skills to plugins:
- On the Core team: https://github.com/microsoft/semantic-kernel/issues/2119
- Added one for our team: https://github.com/microsoft/chat-copilot/issues/86

webapi/CopilotChat/Skills/OpenApiPlugins/README.md
teresaqhoang: Renamed header

webapi/CopilotChat/Skills/OpenApiPlugins/README.md
teresaqhoang: I think this was a copy paste error way back when, we import the Klarna Plugin manifest directly from a URL -- updated this section here to reflect https://www.klarna.com/.well-known/ai-plugin.json

glahaye (approved):

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
teresaqhoang: Actually, I changed it to throw to be more accurate, and formatter confirmed that we should ever run into that case because any exception will throw and propagate up so removing.


image

dehoward: gotcha, so the Planner is guaranteed to return a valid plan at some point?

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
dehoward: ok that makes sense. right now the fn will continue to retry until a valid plan is reached, no matter the value of `maxRetries`, is that intentional?

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
teresaqhoang: It's either guaranteed to return a plan object (not necessarily valid) or throw an error

webapi/CopilotChat/Skills/ChatSkills/ExternalInformationSkill.cs
teresaqhoang: It will throw and consequently bail out if it fails the retry

teresaqhoang: I added the throw back in explicitly

dehoward (approved):

PR #82, opened 2023-08-01
Service option controller

TaoChenOSU:
### Motivation and Context

We would like to show in the webapp some of the AI service options/configurations we select for our webapi.

### Description
1. Add an OptionsController to allow the webapp to pull (non-secretive) configurations from the backend.
2. Webapp to retrieve memories storage option from the webapi and show it in the documents tab.
image

### Contribution Checklist


webapi/CopilotChat/Controllers/OptionsController.cs
```diff
@@ -0,0 +1,41 @@
...
+using SemanticKernel.Service.Options;
+
+namespace SemanticKernel.Service.CopilotChat.Controllers;
+
+/// <summary>
+/// Controller responsible for returning the service options to the client.
+/// </summary>
+[ApiController]
+[Authorize]
+public class OptionsController : ControllerBase
```
dehoward, line 17: let's be more explicit in the naming?
```suggestion
public class ServiceOptionsController : ControllerBase
```

webapp/src/Constants.ts
```diff
@@ -50,4 +50,12 @@ export const Constants = {
         MANIFEST_PATH: '/.well-known/ai-plugin.json',
     },
     KEYSTROKE_DEBOUNCE_TIME_MS: 250,
+    MemoriesStoreTypes: {
```
dehoward, line 53: nit: let's align with the naming in webapi here?
```suggestion
    MemoryStoreType: {
````

File: webapp/src/redux/features/app/appSlice.ts [hunk 3]
```diff
@@ -59,6 +61,22 @@ export const appSlice = createSlice({
...
+        setMemoriesStoreType: (state: AppState, action: PayloadAction<string>) => {
+            const storeType = action.payload;
+            if (Constants.MemoriesStoreTypes.hasOwnProperty(storeType)) {
+                state.memoriesStoreType = action.payload;
+            } else {
+                state.memoriesStoreType = Constants.MemoriesStoreTypes.Unknown;
+
+                const supportedStoreTypes = Object.values(Constants.MemoriesStoreTypes).join(', ');
+                const errorMessage = `Unknown memory store type: ${storeType}. Supported types are: ${supportedStoreTypes}.`;
+                console.log(errorMessage);
```
dehoward, line 73: remove?

webapp/src/redux/features/app/appSlice.ts
```diff
@@ -59,6 +61,22 @@ export const appSlice = createSlice({
                 },
             };
         },
+        setMemoriesStoreType: (state: AppState, action: PayloadAction<string>) => {
+            const storeType = action.payload;
+            if (Constants.MemoriesStoreTypes.hasOwnProperty(storeType)) {
```
dehoward, line 66:
```suggestion
            if (storeType in Constants.MemoryStoreTypes) {
```

webapp/src/App.tsx
```diff
@@ -94,6 +94,9 @@ const App: FC = () => {
                         setAppState(AppState.Chat);
                     }
                 });
+
+                // Load service options
+                void chat.getMemoriesStoreType();
```
dehoward, line 99: looks like we can do this in parallel with the call above:

```suggestion
                void Promise.all([
                    chat.loadChats().then((succeeded) => {
                        if (succeeded) {
                            setAppState(AppState.Chat);
                        }
                    }),
                    chat.getMemoriesStoreType()
                ]);
```

webapp/src/Constants.ts
TaoChenOSU: The webapi uses plural as well.

dehoward: ah I was looking at `GetMemoryType()` in `OptionsController` before, but I see it's changed. imo we should align on `memoryStoryType`, but I don't really have a strong opinion on that. I think alignment around something would be good tho. right now we use `MemoriesStoreTypes` here, `getMemoriesStoreType` in `useChat`, and `memoryStoreType` in the new route.

dehoward: ahhh ok, I see `MemoriesStoreType` defined on the backend now... I think I missed it before. ignore 😄

webapp/src/Constants.ts
```diff
@@ -50,4 +50,12 @@ export const Constants = {
...
+        Qdrant: 'Qdrant',
+        AzureCognitiveSearch: 'AzureCognitiveSearch',
+        Pinecone: 'Pinecone',
+        Milvus: 'Milvus',
+        Volatile: 'Volatile',
+        Unknown: 'Unknown',
```
dehoward, line 59: these seem like values we could/should pull from the backend? some of these options aren't currently supported (e.g. Pinecone & Milvus) and I think may confuse users. 

could we send the keys of `MemoriesStoreOptions.MemoriesStoreType` (or do something else) to communicate to the frontend what types are supported?

TaoChenOSU: I totally agree with what you suggest. In fact, I was thinking of implementing this the way you suggested.
However, I think this is something that we should have inputs from the team. On one hand, we confuse the users by showing things that we don't support (yet?). On the other hand, we want to show people that they can have all those options (in the future?).

Maybe we can create issues and mention the issues in some TODOs as we slowly add supports?

dehoward: let's bring this up in standup tmrw?

I definitely hear your point on advising the user of their options, even if we don't currently support them, I'm just not sure disabled radio buttons is the best way to do that since imo it can bloat the UX with features that we don't have a clear timeline of implementing. 
we may be able to get away with having help text that is either always visible or shown through an icon bubble, with maybe a "learn more" link as well indicating what is on the roadmap storage-wise.. not sure, Ruokan may also have opinions here.

pulling the options dynamically from the backend would allow folks to add other storage types (e.g. Chroma) without making changes to the UX code, which they may be unfamiliar with.

webapi/CopilotChat/Controllers/ServiceOptionsController.cs
```diff
@@ -0,0 +1,49 @@
...
+        this._memoriesStoreOptions = memoriesStoreOptions.Value;
+    }
+
+    /// <summary>
+    /// Return the memory store type that is configured.
+    /// </summary>
+    [Route("memoriesStoreType")]
+    [HttpGet]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public IActionResult GetMemoriesStoreType()
```
teresaqhoang, line 39: Do you foresee us having to expose any other options from the backend (i.e., AI service, model used, etc.)? 

If so, this might make more sense as a GetServiceOptions call that returns a dictionary of KV pairings so we don't have to implement a getter for each Service Option we may need in the future. 

Also not blocking for this PR, we can change this in the future.

webapi/CopilotChat/Models/ServiceOptionsResponse.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json.Serialization;
+
+namespace SemanticKernel.Service.CopilotChat.Models;
+
+/// <summary>
+/// Response to memoriesStoreType request.
+/// </summary>
+public class MemoriesStoreTypeResponse
```
teresaqhoang, line 12: nit the file should be named MissingFunctionErrorOptions.cs to match the class

webapi/CopilotChat/Models/ServiceOptionsResponse.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
+    /// All the available memories store types.
+    /// </summary>
+    [JsonPropertyName("types")]
+    public IEnumerable<string> Types { get; set; } = Enumerable.Empty<string>();
+
+    /// <summary>
+    /// the selected memories store type.
```
dehoward, line 21: super nit: capital 
```suggestion
    /// The selected memories store type.
```

webapi/CopilotChat/Controllers/ServiceOptionsController.cs
TaoChenOSU: I do foresee that, but we don't have a plan yet as what we need to include. And making this complicated will leave little room for design choices down the road unless we reimplement everything. I think we should keep this simple now so that it won't cause us much either way (i.e. one API per option v.s. one API for all options).

teresaqhoang: Yeah sounds good, can you create an issue to track this? and link the number here? So others know not to keep adding getters for each option type

webapi/CopilotChat/Models/ServiceOptionsResponse.cs
TaoChenOSU: We can group other service option responses here if we choose to have multiple APIs or we can define a one-piece service option response that includes all non-secretive options.

webapi/CopilotChat/Controllers/ServiceOptionsController.cs
dehoward: +1 on using one API, instead of calling each API one-by-one when the app loads. 
I don't think the solution needs to be complicated and could be along the lines of:
```csharp
public IActionResult GetServiceOptions()
{
    return this.Ok(new
    {
        memoryStore = new
        {
            Types = Enum.GetNames(typeof(MemoriesStoreOptions.MemoriesStoreType)),
            SelectedType = this._memoriesStoreOptions.Type.ToString()
        }
    });
}
```


 it would be nice to go ahead and set apattern now, so that later, like you mentioned @TaoChenOSU, we don't have to reimplement things down the line, but I also don't want to block the PR more than I already have 😄 

if we don't end up doing it in this PR, can we instead just add a `// Note:` instead of creating an entire issue, since it won't currently be actionable?

webapi/CopilotChat/Controllers/ServiceOptionsController.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    public ServiceOptionsController(
+        ILogger<ServiceOptionsController> logger,
+        IOptions<MemoriesStoreOptions> memoriesStoreOptions)
+    {
+        this._logger = logger;
+        this._memoriesStoreOptions = memoriesStoreOptions.Value;
+    }
+
+    // TODO: [Issue #95] Include all service options in a single response.
```
dehoward, line 33: I'm not sure that we should create an issue for this since it's not a current problem. we want to use a single response whenever we add another option, but we don't know when that will be, and I don't think we should bloat the issues list with things that don't currently need a solution.

dehoward (dismissed):

alliscode (approved):

webapi/CopilotChat/Controllers/ServiceOptionsController.cs
dehoward: synced offline - we'll update this set a pattern and send a bag of options.

webapi/CopilotChat/Controllers/ServiceOptionsController.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    [Route("serviceOptions")]
+    [HttpGet]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public IActionResult GetServiceOptions()
+    {
+        return this.Ok(
+            new ServiceOptionsResponse()
+            {
+                MemoriesStoreType = new MemoriesStoreTypeResponse()
```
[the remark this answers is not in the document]
TaoChenOSU: Good catch!

dehoward (approved):

teresaqhoang (approved):

PR #83, opened 2023-08-01
Upload documents in parallel + truncate text for long document names

### LINKED ISSUE #49: Import documents in parallel if > 1
If multiple docs are uploaded, import in parallel

dehoward:
### Motivation and Context

resolves #49.

### Description 

- `DocumentImportController`: changed to use `Task.WhenAll` when importing documents.
- all other files: changed to show overflow text on hover and to truncate long document names (shown below).
  image
  image
  image

### Contribution Checklist

### DIFF
--- a/webapi/CopilotChat/Controllers/DocumentImportController.cs
+++ b/webapi/CopilotChat/Controllers/DocumentImportController.cs
@@ -126,18 +126,21 @@ public class DocumentImportController : ControllerBase
 
         this._logger.LogInformation("Importing {0} document(s)...", documentImportForm.FormFiles.Count());
 
-        // TODO: [Issue #49] Perform the import in parallel.
         DocumentMessageContent documentMessageContent = new();
         IEnumerable<ImportResult> importResults = new List<ImportResult>();
-        foreach (var formFile in documentImportForm.FormFiles)
-        {
-            var importResult = await this.ImportDocumentHelperAsync(kernel, formFile, documentImportForm);
-            documentMessageContent.AddDocument(
-                formFile.FileName,
-                this.GetReadableByteString(formFile.Length),
-                importResult.IsSuccessful);
-            importResults = importResults.Append(importResult);
-        }
+        await Task.WhenAll(documentImportForm.FormFiles.Select(formFile =>
+            this.ImportDocumentHelperAsync(kernel, formFile, documentImportForm).ContinueWith(task =>
+                {
+                    var importResult = task.Result;
+                    if (importResult != null)
+                    {
+                        documentMessageContent.AddDocument(
+                            formFile.FileName,
+                            this.GetReadableByteString(formFile.Length),
+                            importResult.IsSuccessful);
+                        importResults = importResults.Append(importResult);
+                    }
+                }, TaskScheduler.Default)));
 
         // Broadcast the document uploaded event to other users.
         if (documentImportForm.DocumentScope == DocumentImportForm.DocumentScopes.Chat)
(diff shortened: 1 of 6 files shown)


webapp/src/components/chat/chat-list/ChatListItem.tsx
```diff
@@ -55,12 +55,9 @@ const useClasses = makeStyles({
         justifyContent: 'space-between',
     },
     title: {
-        ...shorthands.overflow('hidden'),
-        textOverflow: 'ellipsis',
-        whiteSpace: 'nowrap',
+        ...SharedStyles.overflowEllipsis,
         fontSize: tokens.fontSizeBase300,
         color: tokens.colorNeutralForeground1,
-        lineHeight: tokens.lineHeightBase200,
```
dehoward, line 63: removed this to fix clipping issue:
image

webapp/src/components/chat/chat-history/ChatHistory.tsx
```diff
@@ -31,17 +31,14 @@ export const ChatHistory: React.FC<ChatHistoryProps> = ({ messages, onGetRespons
 
     return (
         <div className={classes.root}>
-            {messages
-                .slice()
-                .sort((a, b) => a.timestamp - b.timestamp)
```
dehoward, line 36: this isn't needed because we already order the messages correctly using the `.reverse()` in `ChatService`

webapi/CopilotChat/Controllers/DocumentImportController.cs
```diff
@@ -126,18 +126,21 @@ private enum SupportedFileType
...
-        foreach (var formFile in documentImportForm.FormFiles)
-        {
-            var importResult = await this.ImportDocumentHelperAsync(kernel, formFile, documentImportForm);
-            documentMessageContent.AddDocument(
-                formFile.FileName,
-                this.GetReadableByteString(formFile.Length),
-                importResult.IsSuccessful);
-            importResults = importResults.Append(importResult);
-        }
+        await Task.WhenAll(documentImportForm.FormFiles.Select(formFile =>
```
TaoChenOSU, line 131: We need to make sure the kernel supports inserting memories in parallel. 
My experience with this is that the VolatileMemoryStore is not made to support inserting memories in parallel. Specially, you could potentially end up getting this exception: https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel/Memory/VolatileMemoryStore.cs#L25.

webapp/src/components/chat/chat-history/ChatHistoryDocumentContent.tsx
```diff
@@ -81,23 +85,30 @@ export const ChatHistoryDocumentContent: React.FC<ChatHistoryDocumentContentProp
 
     return (
         <>
-            {documents.map((document, index) => (
+            {documents.map(({ name, size, isUploaded }, index) => (
```
teresaqhoang, line 88: nit: this would keep things better defined
```suggestion
            {documents.map((document, index) => (
            const { name, size, isUploaded } = document;
```

teresaqhoang (commented):
Will approve after Tao's comment is resolved.

webapp/src/components/chat/chat-history/ChatHistoryDocumentContent.tsx
dehoward: I tend to prefer avoiding multiple accesses of the same property by destructuring the variable, esp when the variable isn't needed. will leave as-is

webapi/CopilotChat/Controllers/DocumentImportController.cs
dehoward: this has merged and the exception should no longer be sent. specifically, when `CreateCollectionAsync` is called during the memory write flow, `VolatileMemoryStore` will no longer throw an exception if the collection already exists.

dehoward: we will need to consume the updated nuget to benefit from this change, but this work should happen soon as we have been doing this regularly.

TaoChenOSU (approved):

PR #92, opened 2023-08-02
Security: remove API key auth

gitri-ms:
### Motivation and Context

This PR addresses https://github.com/microsoft/semantic-kernel/issues/1639

### Description

1. Remove API key authorization
2. Use "AzureAD" as default authentication configuration for deployments, "None" for running locally _(**Note:** UI changes to disable sign in flow for the latter case are still forthcoming)_
4. Enable auth policy on controllers that checks if the user is part of the conversation they are trying to access
5. Update routes to better align to Microsoft API design principles

**Breaking change:** This PR changes the contract between the frontend and backend around how user IDs are communicated. Users who have been signing into the frontend with AAD will now only see their chats if the backend is also gated by AAD authentication, which was not the case previously.

### Contribution Checklist


deploy/README.md
```diff
@@ -112,8 +112,6 @@ This will get you to the CORS page where you can add your allowed hosts.
 
 ## Authorization
 All of endpoints (except `/healthz`) require authorization to access.
-By default, an API key is required for access which can be found in the `Authorization:ApiKey` configuration setting.
```
gitri-ms, line 115: TODO: update this section

deploy/deploy-webapp.ps1
```diff
@@ -24,10 +24,10 @@ param(
     # Client application id
     $ApplicationClientId,
 
-    [Parameter(Mandatory=$false)]
+    [Parameter(Mandatory = $false)]
     [string]
     # Authority for client applications that are not configured as multi-tenant.
-    $Authority="https://login.microsoftonline.com/common"
+    $Authority = "https://login.microsoftonline.com/common"
```
gitri-ms, line 30: There is some weirdness here -- the backend expects instance + tenant ID, while the frontend expects authority. However the values need to match. This could be confusing for the user. Perhaps change to make them more consistent (e.g. both use instance + tenant id)?

webapi/Auth/AuthInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+            var tenantIdClaim = user.FindFirst(ClaimConstants.Tid)
+                ?? user.FindFirst(ClaimConstants.TenantId);
+            var userNameClaim = user.FindFirst(ClaimConstants.Name);
+            if (userNameClaim is null)
+            {
+                throw new CredentialUnavailableException("User name was not present in the request token.");
+            }
+            return new AuthData
+            {
+                UserId = (tenantIdClaim is null) ? userIdClaim.Value : string.Join(".", userIdClaim.Value, tenantIdClaim.Value),
```
gitri-ms, line 47: Alternatively can make tenant id a requirement and provide a dummy tenant id for the default user. Need to better understand the implications on the front end -- what happens if there is no tenant id?

deploy/deploy-azure.ps1
```diff
@@ -44,6 +49,14 @@ param(
     # SKU for the Azure App Service plan
     $WebAppServiceSku = "B1",
 
+    [string]
+    # Azure AD cloud instance for authenticating users
+    $AzureAdInstance = "https://login.microsoftonline.com/",
+
+    [string]
+    # Azure AD tenant ID for authenticating users
+    $AzureAdTenantId = "common",
```
glahaye, line 58: As discussed during bug bash, we might want to not put "common" as the default and force users to provide a tenant

deploy/deploy-azure.sh
```diff
@@ -5,19 +5,22 @@
...
     echo "  -aikey, --ai-service-key AI_SERVICE_KEY    API key for existing Azure OpenAI resource or OpenAI account"
     echo "  -aiend, --ai-endpoint AI_ENDPOINT          Endpoint for existing Azure OpenAI resource"
     echo "  -rg, --resource-group RESOURCE_GROUP       Resource group to which to make the deployment (default: \"rg-\$DEPLOYMENT_NAME\")"
     echo "  -r, --region REGION                        Region to which to make the deployment (default: \"South Central US\")"
     echo "  -wr, --web-app-region WEB_APP_REGION       Region to deploy to the static web app into. This must be a region that supports static web apps. (default: \"West US 2\")"
     echo "  -a, --app-service-sku WEB_APP_SVC_SKU      SKU for the Azure App Service plan (default: \"B1\")"
+    echo "  -i, --instance AZURE_AD_INSTANCE           Azure AD cloud instance for authenticating users"
+    echo "                                             (default: \"https://login.microsoftonline.com/\")"
+    echo "  -t, --tenant-id AZURE_AD_TENANT_ID         Azure AD tenant ID for authenticating users (default: \"common\")"
```
glahaye, line 23: Again, might want to make this mandatory...

deploy/deploy-azure.sh
```diff
@@ -171,20 +179,24 @@ az account set -s "$SUBSCRIPTION"
 : "${REGION:="southcentralus"}"
 : "${WEB_APP_SVC_SKU:="B1"}"
 : "${WEB_APP_REGION:="westus2"}"
+: "${AZURE_AD_INSTANCE:="https://login.microsoftonline.com/"}"
+: "${AZURE_AD_TENANT_ID:="common"}"
```
glahaye, line 183: Again...

deploy/deploy-webapp.ps1
glahaye: What's the format expected from the client side?

deploy/deploy-azure.ps1
gitri-ms: Yup, I am currently going through and changing all of these

deploy/main.bicep
```diff
@@ -39,9 +39,14 @@ param aiEndpoint string = ''
...
-@description('WebAPI key to use for authorization')
-param webApiKey string = newGuid()
+@description('Azure AD client ID for the backend web API')
+param webApiClientId string = ''
+
+@description('Azure AD cloud instance for authenticating users')
+param azureAdInstance string = environment().authentication.loginEndpoint
+
+@description('Azure AD tenant ID for authenticating users')
+param azureAdTenantId string = 'common'
```
glahaye, line 49: Default again...

gitri-ms:
Closing this PR as it is now covered by #126

PR #110, opened 2023-08-05
Change AAD authorization to be opt-in within webapp

dehoward:
### Description

dependent on https://github.com/microsoft/chat-copilot/pull/92. 
removing AAD as the default authorization method will allow users to get the app running more quickly. 

**_breaking change:_** existing users will need to uncomment `REACT_APP_AUTH_TYPE=AzureAd` in `webapp/.env` to continue using AAD as their authorization type.

### Contribution Checklist
- adds `REACT_APP_AUTH_TYPE` and changes AAD variables in `.env` to be optional 
- adds `AuthHelper.IsAuthAAD` to conditionally render different elements throughout the app
- changes user settings menu popup to instead just show as a settings button:
image

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [x] All unit tests pass, and I have added new tests where possible
- [ ] I didn't break anyone :smile:

PR #126, opened 2023-08-08
Chat Copilot Security Improvements

gitri-ms:
### Motivation and Context
This PR addresses https://github.com/microsoft/semantic-kernel/issues/1639. It is a combination of PRs #92 and #110 

#### Backend changes
- Remove API key authorization
- Use "AzureAD" as default authentication configuration for deployments, "None" for running locally (Note: UI changes to disable sign in flow for the latter case are still forthcoming)
- Enable auth policy on controllers that checks if the user is part of the conversation they are trying to access

This PR changes the contract between the frontend and backend around how user IDs are communicated. Users who have been signing into the frontend with AAD will now only see their chats if the backend is also gated by AAD authentication, which was not the case previously.

#### Frontend changes
- adds `REACT_APP_AUTH_TYPE` and changes AAD variables in `.env` to be optional 
- adds `AuthHelper.IsAuthAAD` to conditionally render different elements throughout the app
- changes user settings menu popup to instead just show as a settings button:
image

Existing users will need to uncomment `REACT_APP_AUTH_TYPE=AzureAd` in `webapp/.env` to continue using AAD as their authorization type.

### Contribution Checklist


deploy/deploy-azure.ps1
```diff
@@ -44,6 +49,14 @@ param(
     # SKU for the Azure App Service plan
     $WebAppServiceSku = "B1",
 
+    [string]
+    # Azure AD cloud instance for authenticating users
+    $AzureAdInstance = "https://login.microsoftonline.com/",
+
+    [string]
+    # Azure AD tenant ID for authenticating users
+    $AzureAdTenantId = "common",
```
gitri-ms, line 58: TODO: update all scripts to require a tenant ID and no longer use "common"

dehoward:
updated the bot's persona to be a generic icon instead of the "DU" initials:
image

deploy/deploy-azure.sh
```diff
@@ -5,20 +5,22 @@
 set -e
 
 usage() {
-    echo "Usage: $0 -d DEPLOYMENT_NAME -s SUBSCRIPTION -ai AI_SERVICE_TYPE -aikey AI_SERVICE_KEY [OPTIONS]"
+    echo "Usage: $0 -d DEPLOYMENT_NAME -s SUBSCRIPTION -c WEBAPI_CLIENT_ID -ai AI_SERVICE_TYPE -aikey AI_SERVICE_KEY [OPTIONS]"
```
glahaye, line 8: I believe AZURE_AD_TENANT_ID should be here too

deploy/deploy-azure.sh
```diff
@@ -109,16 +121,11 @@ while [[ $# -gt 0 ]]; do
 done
 
 # Check mandatory arguments
-if [[ -z "$DEPLOYMENT_NAME" ]] || [[ -z "$SUBSCRIPTION" ]] || [[ -z "$AI_SERVICE_TYPE" ]]; then
+if [[ -z "$DEPLOYMENT_NAME" ]] || [[ -z "$SUBSCRIPTION" ]] || [[ -z "$WEBAPI_CLIENT_ID" ]] || [[ -z "$AI_SERVICE_TYPE" ]]; then
```
glahaye, line 124: What about AZURE_AD_TENANT_ID?

README.md
```diff
@@ -46,19 +39,12 @@ You will need the following items to run the sample:
 3. Configure Chat Copilot.
   
     ```powershell
-    .\Configure.ps1 -AIService {AI_SERVICE} -APIKey {API_KEY} -Endpoint {AZURE_OPENAI_ENDPOINT} -ClientId {AZURE_APPLICATION_ID} 
+    .\Configure.ps1 -AIService {AI_SERVICE} -APIKey {API_KEY} -Endpoint {AZURE_OPENAI_ENDPOINT}
```
gitri-ms, line 42: Need to update local scripts to take client/tenant/instance as optional params

webapi/Controllers/DocumentImportController.cs
```diff
@@ -101,12 +103,12 @@ private enum SupportedFileType
...
         this._participantRepository = participantRepository;
         this._ocrEngine = ocrEngine;
+        this._authInfo = authInfo;
     }
 
     /// <summary>
     /// Service API for importing a document.
     /// </summary>
-    [Authorize]
     [Route("importDocuments")]
```
gitri-ms, line 112: Route: "documents"

webapi/Controllers/BotController.cs
```diff
@@ -153,21 +151,21 @@ public class BotController : ControllerBase
     /// <param name="kernel">The Semantic Kernel instance.</param>
     /// <param name="chatId">The chat id to be downloaded.</param>
     /// <returns>The serialized Bot object of the chat id.</returns>
-    [Authorize]
     [HttpGet]
     [ActionName("DownloadAsync")]
     [Route("bot/download/{chatId:guid}")]
```
gitri-ms, line 156: Route: bots/<chatId>

webapi/Controllers/BotController.cs
```diff
@@ -73,23 +73,21 @@ public class BotController : ControllerBase
...
     /// </summary>
     /// <param name="kernel">The Semantic Kernel instance.</param>
-    /// <param name="userId">The user id.</param>
+    /// <param name="authInfo">The auth info instance.</param>
     /// <param name="bot">The bot object from the message body</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>The HTTP action result with new chat session object.</returns>
-    [Authorize]
     [HttpPost]
     [Route("bot/upload")]
```
gitri-ms, line 81: Route: "bots"

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -69,18 +73,19 @@ public class ChatHistoryController : ControllerBase
     /// <param name="chatParameter">Contains the title of the chat.</param>
     /// <returns>The HTTP action result.</returns>
     [HttpPost]
-    [Route("chatSession/create")]
+    [Route("chatSessions")]
```
gitri-ms, line 76: Revert all route changes - will be done in a subsequent PR

webapi/Controllers/ChatController.cs
```diff
@@ -70,17 +75,40 @@ public ChatController(ILogger<ChatController> logger, ITelemetryService telemetr
...
+        [FromServices] IAuthInfo authInfo,
+        [FromBody] ChatAsk ask)
     {
         this._logger.LogDebug("Chat request received.");
 
-        // Put ask's variables in the context we will use.
-        var contextVariables = new ContextVariables(ask.Input);
-        foreach (var input in ask.Variables)
+        // Verify that the chat exists and that the user has access to it.
+        const string chatIdKey = "chatId";
```
glahaye, line 87: Nit: constants in C# use PascalCase, not camelCase (chatIdKey -> ChatIdKey)

webapi/Controllers/ChatController.cs
```diff
@@ -70,17 +75,40 @@ public ChatController(ILogger<ChatController> logger, ITelemetryService telemetr
...
+        var chatIdFromContext = ask.Variables.FirstOrDefault(x => x.Key == chatIdKey);
+        if (chatIdFromContext.Key is chatIdKey)
```
glahaye, line 89: This will explode if you get null on the previous line.

```suggestion
        if (chatIdFromContext?.Key is chatIdKey)
```

webapi/Controllers/ChatController.cs
```diff
@@ -58,9 +60,12 @@ public ChatController(ILogger<ChatController> logger, ITelemetryService telemetr
...
+    /// <param name="askConverter">Converter to use for converting Asks.</param>
+    /// <param name="chatSessionRepository">Repository of chat sessions.</param>
+    /// <param name="chatParticipantRepository">Repository of chat participants.</param>
+    /// <param name="authInfo">Auth info for the current request.</param>
     /// <param name="ask">Prompt along with its parameters.</param>
     /// <returns>Results containing the response from the model.</returns>
-    [Authorize]
     [Route("chat")]
     [HttpPost]
     [ProducesResponseType(StatusCodes.Status200OK)]
```
glahaye, line 71: Is also produces 404, and 403...

webapi/Controllers/SpeechTokenController.cs
```diff
@@ -6,15 +6,13 @@
...
 using CopilotChat.WebApi.Options;
-using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace CopilotChat.WebApi.Controllers;
 
-[Authorize]
```
glahaye, line 17: How is this gated now?  What prevents any random dude from obtaining speech service tokens from us now?

webapi/Models/Request/CreateChatParameters.cs
```diff
@@ -1,24 +1,15 @@
...
 
-using System.Text.Json.Serialization;
-
 namespace CopilotChat.WebApi.Models.Request;
 
 /// <summary>
-/// Json body for creating a new chat session.
+/// Parameters for creating a new chat session.
 /// </summary>
 public class CreateChatParameters
```
glahaye, line 9: Please remove this empty line

webapi/Models/Storage/ChatMessage.cs
```diff
@@ -15,6 +14,8 @@ namespace CopilotChat.WebApi.Models.Storage;
 /// </summary>
 public class ChatMessage : IStorageEntity
 {
+    private static readonly JsonSerializerOptions SerializerSettings = new() { PropertyNamingPolicy = JsonNamingPolicy.CamelCase };
```
glahaye, line 17: Cool!

webapp/.env.example
```diff
@@ -1,15 +1,14 @@
...
 
-# Optional Variables
-REACT_APP_SK_API_KEY=
-
-# Authorization scopes to access webapi when using Azure AD authorization.
+# To enable authorization using Azure Active Directory, uncomment the following variables
 # See paragraph "(Optional) Enable backend authorization via Azure AD" in README.md for details and setup
-REACT_APP_AAD_API_SCOPE=
+# REACT_APP_AUTH_TYPE=AzureAd
+# REACT_APP_AAD_AUTHORITY=https://login.microsoftonline.com/common
```
glahaye, line 8: We mention "common" here...  Shouldn't we say something like [YOUR_TENANT_ID] instead?

README.md
dehoward: and to set `REACT_APP_AUTH_TYPE=AzureAd` in `.env` if these params are passed

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: Why? I'm vibing the new routes

webapp/src/index.tsx
```diff
@@ -44,18 +45,30 @@ document.addEventListener('DOMContentLoaded', () => {
...
-                        </MsalProvider>
+                        {/* eslint-disable @typescript-eslint/no-non-null-assertion */}
+                        {shouldUseMsal ? (
+                            <MsalProvider instance={msalInstance!}>
+                                <AppWithTheme />
+                            </MsalProvider>
+                        ) : (
+                            <AppWithTheme />
+                        )}
+                        {/* eslint-enable @typescript-eslint/no-non-null-assertion */}
```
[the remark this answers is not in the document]
dehoward: this is needed, notice that it's an `eslint-enable` comment instead of a disable. it's because `eslint-disable-next-line` cannot be used above since the comment is not allowed inside the ternary.

webapp/src/libs/auth/AuthHelper.ts
```diff
@@ -9,13 +9,34 @@ import {
...
 import { TokenHelper } from './TokenHelper';
 
 const log = debug(Constants.debug.root).extend('authHelper');
 
+export const enum AuthType {
+    AAD = 'AzureAd',
+}
+
+export const DefaultChatUser: IChatUser = {
+    id: 'c05c61eb-65e4-4223-915a-fe72b0c9ece1',
```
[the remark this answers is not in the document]
dehoward: [from the backend](https://github.com/microsoft/chat-copilot/pull/126/files#diff-7b039d0ec1e1d963a9f72ca782023aa20a7b6ed4c471879eac6dc8d96a9b5b26R19). it's the same ID being used there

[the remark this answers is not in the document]
gitri-ms: Yup it's just a random GUID. It can be any GUID, just has to match on both sides.

webapi/Models/Request/Ask.cs
```diff
@@ -7,7 +7,9 @@ namespace CopilotChat.WebApi.Models.Request;
 
 public class Ask
 {
-    public string Input { get; set; } = string.Empty;
+    public virtual string Input { get; set; } = string.Empty;
 
     public IEnumerable<KeyValuePair<string, string>> Variables { get; set; } = Enumerable.Empty<KeyValuePair<string, string>>();
 }
+
+// TODO: Does it make sense to keep this here? Or should combine Ask and ChatAsk into one, under CopilotChat/Models?
```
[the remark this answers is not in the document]
gitri-ms: Doesn't look like they are used separately. I think it was an artifact of the prior folder structure, with CopilotChat being a subset of WebApi. I am merging them.

webapi/Controllers/ChatHistoryController.cs
gitri-ms: Because it's a breaking change that customers need to be aware of. We need to provide them with a way to transition without downtime. We need to ensure that we make such changes only once (do it right) and thoughtfully. But we don't want to hold the rest of the security improvements until then.

webapi/Controllers/ChatController.cs
gitri-ms: `chatIdFromContext` is of type `KeyValuePair<string, string>` -- it doesn't look like it can be null?

webapi/Controllers/SpeechTokenController.cs
gitri-ms: All controllers have authorization turned on by default, per the code changes in `ServiceExtensions.cs`

webapi/Utilities/AskConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// <summary>
+    /// Converts <see cref="Ask"/> variables to <see cref="ContextVariables"/>, inserting some system variables along the way.
+    /// </summary>
+    public ContextVariables GetContextVariables(Ask ask)
+    {
+        const string userIdKey = "userId";
+        const string userNameKey = "userName";
+        var contextVariables = new ContextVariables(ask.Input);
+        foreach (var input in ask.Variables)
```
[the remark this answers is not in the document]
glahaye: Or variable instead of simply "var"

webapp/src/libs/auth/AuthHelper.ts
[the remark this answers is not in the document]
alliscode: It might help to add a comment explaining that the default user is what is used when auth is set to None, and that the user id just needs to match what the backend expects.

webapi/Controllers/ChatParticipantController.cs
```diff
@@ -86,11 +88,12 @@ public class ChatParticipantController : ControllerBase
     /// <summary>
     /// Get a list of chat participants that have the same chat id.
     /// </summary>
-    /// <param name="chatId">The Id of the chat to get all the participants from.</param>
+    /// <param name="chatId">The ID of the chat to get all the participants from.</param>
     [HttpGet]
-    [Route("chatParticipant/getAllParticipants/{chatId:guid}")]
+    [Route("chatSessions/{chatId:guid}/participants")]
```
gitri-ms, line 93: change back

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -122,22 +129,25 @@ public async Task<IActionResult> GetChatSessionByIdAsync(Guid chatId)
     }
 
     /// <summary>
-    /// Get all chat sessions associated with a user. Return an empty list if no chats are found.
-    /// The regex pattern that is used to match the user id will match the following format:
-    ///    - 2 period separated groups of one or more hyphen-delimited alphanumeric strings.
-    /// The pattern matches two GUIDs in canonical textual representation separated by a period.
+    /// Get all chat sessions associated with the logged in user. Return an empty list if no chats are found.
     /// </summary>
-    /// <param name="userId">The user id.</param>
```
gitri-ms, line 130: put back

README.md
```diff
@@ -134,17 +113,27 @@ You will need the following items to run the sample:
...
-   > For more details on creating an application registration, go [here](https://learn.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app).
+### Requirements
+
+- [Azure account](https://azure.microsoft.com/free)
+- [Azure AD Tenant](https://learn.microsoft.com/azure/active-directory/develop/quickstart-create-new-tenant)
+
+### Instructions
+
+1. Create an [application registration](https://learn.microsoft.com/azure/active-directory/develop/quickstart-register-app) for the frontend web app, using the values below
+    - `Supported account types`: "_Accounts in any organizational directory (Any Azure AD directory - Multitenant) and personal Microsoft accounts (e.g. Skype, Xbox)_"
```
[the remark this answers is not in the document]
gitri-ms: yes, should definitely do this. i'll update.

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -150,9 +160,9 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
             else
             {
                 this._logger.LogDebug(
-                    "Failed to find chat session with id {0} for participant {1}", chatParticipant.ChatId, chatParticipant.Id);
+                    "Failed to find chat session with id {0}", chatParticipant.ChatId);
                 return this.NotFound(
```
alliscode, line 164: I don't think a 404 is the correct response here. If I understand this correctly, then the case here is that we found a matching `chatParticipant` record for the user, but cannot find the matching `chatSession`.

- If this is considered a valid case (sems weird), then we should return an empty collection to match the expected return type when no chats are found for the user.
- If this is not a valid case, and it indicated that something is corrupted or out of sync, then a 500 level error would make more sense.

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -122,22 +129,25 @@ public async Task<IActionResult> GetChatSessionByIdAsync(Guid chatId)
...
-    /// Get all chat sessions associated with a user. Return an empty list if no chats are found.
-    /// The regex pattern that is used to match the user id will match the following format:
-    ///    - 2 period separated groups of one or more hyphen-delimited alphanumeric strings.
-    /// The pattern matches two GUIDs in canonical textual representation separated by a period.
+    /// Get all chat sessions associated with the logged in user. Return an empty list if no chats are found.
     /// </summary>
-    /// <param name="userId">The user id.</param>
     /// <returns>A list of chat sessions. An empty list if the user is not in any chat session.</returns>
     [HttpGet]
     [Route("chatSession/getAllChats/{userId:regex(([[a-z0-9]]+-)+[[a-z0-9]]+\\.([[a-z0-9]]+-)+[[a-z0-9]]+)}")]
```
alliscode, line 136: This regex will fail to match in the no auth case since the userId is a single GUID instead of user.tenant. I think we can just remove the regex contraint for now and allow the {userId} to be any value since we can ignore it anyway.

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -122,22 +129,25 @@ public async Task<IActionResult> GetChatSessionByIdAsync(Guid chatId)
...
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
     {
+        // TODO: [Issue #141] Remove this once we remove userId from route
+        if (!userId.Equals(this._authInfo.UserId, StringComparison.Ordinal))
+        {
+            return this.Forbid("User id does not match request.");
+        }
+
```
alliscode, line 147: I think this can be removed. We use the userId that comes from the auth header value anyway so the userId in the route is just a temporary placeholder to keep things backward compatible. We don't really care what the value is.

webapi/Controllers/ChatHistoryController.cs
gitri-ms: I'm not sure which is better here. It doesn't seem like a valid case, but should we still return any chats we do find? I could envision us possibly getting into an out of sync state once something like delete chat is implemented.

scripts/deploy/README.md
```diff
@@ -11,24 +11,81 @@ This document details how to deploy CopilotChat's required resources to your Azu
...
+## App registrations (identity)
+
+You will need two Azure Active Directory (AAD) application registrations -- one for the frontend web app and one for the backend API.
+
+> For details on creating an application registration, go [here](https://learn.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app).
+
+### Frontend app registration
+
+- Select `Single-page application (SPA)` as platform type, and set the redirect URI to `http://localhost:3000`
+- Select `Accounts in any organizational directory and personal Microsoft Accounts` as supported account types.
```
gitri-ms, line 36: Change guidance to single tenant (and below)

scripts/deploy/deploy-webapp.ps1
```diff
@@ -24,10 +24,10 @@ param(
     # Client application id
     $ApplicationClientId,
 
-    [Parameter(Mandatory=$false)]
+    [Parameter(Mandatory = $false)]
     [string]
     # Authority for client applications that are not configured as multi-tenant.
-    $Authority="https://login.microsoftonline.com/common"
+    $Authority = "https://login.microsoftonline.com/common"
```
gitri-ms, line 30: Use tenant instead of common

scripts/deploy/deploy-webapp.sh
```diff
@@ -7,14 +7,15 @@ set -e
...
     echo ""
     echo "Arguments:"
     echo "  -s, --subscription SUBSCRIPTION        Subscription to which to make the deployment (mandatory)"
     echo "  -rg, --resource-group RESOURCE_GROUP   Resource group name from a 'deploy-azure.sh' deployment (mandatory)"
     echo "  -d, --deployment-name DEPLOYMENT_NAME  Name of the deployment from a 'deploy-azure.sh' deployment (mandatory)"
-    echo "  -a, --application-id APPLICATION_ID    Client application ID (mandatory)"
-    echo "  -au, --authority                       Authority to use for client applications that are not configured as multi-tenant. Defaults to (https://login.microsoftonline.com/common) if not specified."
+    echo "  -c, --client-id FRONTEND_CLIENT_ID     Client application ID for the frontend web app (mandatory)"
+    echo "  -t, --tenant-id AZURE_AD_TENANT_ID     Azure AD tenant ID (mandatory)"
+    echo "  -i, --instance AZURE_AD_INSTANCE       Azure cloud instance for authenticating users. Defaults to (https://login.microsoftonline.com/) if not specified."
```
gitri-ms, line 18: Need to update workflows to reflect these changes

scripts/deploy/deploy-webapp.ps1
```diff
@@ -21,13 +21,18 @@ param(
...
+    [string]
+    # Azure AD tenant id
+    $TenantId,
+
+    [Parameter(Mandatory = $false)]
     [string]
-    # Authority for client applications that are not configured as multi-tenant.
-    $Authority="https://login.microsoftonline.com/common"
+    # Azure cloud instance for authenticating users
+    $Instance = "https://login.microsoftonline.com"
```
gitri-ms, line 35: Can pull these from az webapp config rather than have them as params

scripts/Configure.sh
```diff
@@ -72,13 +82,26 @@ fi
...
 
+if [ "$FRONTEND_CLIENT_ID" ] && [ "$BACKEND_CLIENT_ID" ] && [ "$TENANT_ID" ]; then
+  # Set auth type to AzureAd
+  AUTH_TYPE="$ENV_AZURE_AD"
+  # If instance empty, use default
+  if [ -z "$INSTANCE" ]; then
+    INSTANCE="$ENV_INSTANCE"
+  fi
+else
+  if [ "$FRONTEND_CLIENT_ID" ] && [ "$BACKEND_CLIENT_ID" ] && [ "$TENANT_ID" ]; then
```
gitri-ms, line 97:
```suggestion
  if [ -z "$FRONTEND_CLIENT_ID" ] && [ -z "$BACKEND_CLIENT_ID" ] && [ -z "$TENANT_ID" ]; then
```

scripts/Start.ps1
```diff
@@ -7,7 +7,7 @@ $BackendScript = Join-Path "$PSScriptRoot" 'Start-Backend.ps1'
 $FrontendScript = Join-Path "$PSScriptRoot" 'Start-Frontend.ps1'
 
 # Start backend (in new PS process)
-Start-Process powershell -ArgumentList "-command $BackendScript"
+Start-Process pwsh -ArgumentList  "-command $BackendScript"
```
gitri-ms, line 10:
```suggestion
Start-Process pwsh -ArgumentList "-command $BackendScript"
```

webapi/appsettings.json
```diff
@@ -223,7 +220,10 @@
   // Filter for hostnames app can bind to
   "AllowedHosts": "*",
   // CORS
-  "AllowedOrigins": ["http://localhost:3000", "https://localhost:3000"],
+  "AllowedOrigins": [
+    "http://localhost:3000",
```
alliscode, line 224: Do we need to enable http?

gitri-ms: This is just a whitespace change (autoformatting from VS Code), no functional changes.

alliscode (approved):

PR #134, opened 2023-08-08
🐛 Fix connection error handling + uncaught exceptions

teresaqhoang:
### Motivation and Context

Fix connection errors in useChat and appSlice + fix red screens of death

### Description

- Added an ID to Alerts in AppState, add logic to update connection status in appSlice, and update SignalRMiddleware to add an ID to connection alerts.
- Added more error handling to reduce red screens of death
- Move all getAccessToken calls into try catch blocks
- Added some missing details around Stepwise Planner

### Contribution Checklist


webapp/src/redux/features/app/appSlice.ts
```diff
@@ -89,3 +94,38 @@ const getTotalTokenUsage = (previousSum?: number, current?: number) => {
...
+};
+
+const updateConnectionStatus = (state: AppState, statusUpdate: Alert) => {
+    if (isServerConnectionError(statusUpdate.message)) {
+        statusUpdate.message =
+            // Constant message so alert UI doesn't feel glitchy on every connection error from SignalR
+            'Cannot send data due to lost connection or server timeout. Try refreshing this page to restart the connection.';
+    }
+
+    // There should only ever be one connection alert at a time,
```
dehoward, line 120: I would go as far to say there only ever be one alert at a time regardless, no mattter what it's message is. 
I think usually the previous alert is irrelevant after the next one anyway, and the UI can get cluttered when we have multiple, especially at 3.

webapp/src/checkEnv.ts
```diff
@@ -1,7 +1,8 @@
+// Checks if all required environment variables are defined
+// Returns an array of missing variables
```
dehoward, line 2: can we use JSDoc comments here?

dehoward (changes requested):

webapp/src/redux/features/app/appSlice.ts
teresaqhoang: Since we allow concurrent chats and updates, alerts can be shown for errors from a different chat, so I think it's important to allow multiple errors in parallel.

I think errors are necessary information and shouldn't be deprioritized over worries of a cluttered UI, especially if they're communicating unexpected behavior in UX. If we limit to one, or even 2, an error can be quickly lost in favor of another before the user can understand what's going on, introducing a race condition in error UX.

teresaqhoang: It makes sense to limit the server connection updates to one alert because they're communicating the state of the same stateful object (connection to the server), so when we remove the existing alert, it's no longer relevant. Which is not the case for other errors

dehoward: > alerts can be shown for errors from a different chat

right, this seems a bit strange. if you're in one chat, you probably shouldn't see errors from another chat, unless that error is relevant to the entire window, like a connection alert or join chat failure. 

Teams has this concept, where some errors are shown in chat:
image

vs. others shown globally:
image

> If we limit to one, or even 2, an error can be quickly lost in favor of another before the user can understand what's going on, introducing a race condition in error UX

ideally, we shouldn't have errors firing so rapidly in the app, that a user cannot keep up. I won't block on this tho, since we currently show 3 and there are no complaints 😄

dehoward (approved):

PR #145, opened 2023-08-09
🚀 Adding new app state to handle user info

teresaqhoang:
### Motivation and Context

Added new app state for setting user info, refactored app views, and added shared styles for informative views.

This is a fix for a bug where the webapp would get stuck on the Loading Chats page. Something with SSO caused this.

### Description

- The App component now transitions to a SettingUserInfo state after successfully discovering the backend, and then to the LoadingChats state after successfully setting the active user info.
- The user info status text in the App component has been updated to display a default message while fetching user information, and an error component has been added to display an error message at view level. 
- A new useSharedClasses hook has been added to provide a set of shared styles for informative views. 
-  The BackendProbe component has also been updated to use the shared informativeView class.
- Added tesseract data files to gitignore

### Contribution Checklist

### DIFF
--- a/webapp/src/App.tsx
+++ b/webapp/src/App.tsx
@@ -48,6 +48,7 @@ export const useClasses = makeStyles({
 enum AppState {
     ProbeForBackend,
     SettingUserInfo,
+    ErrorLoadingUserInfo,
     LoadingChats,
     Chat,
     SigningOut,
@@ -62,8 +63,6 @@ const App: FC = () => {
     const { instance, inProgress } = useMsal();
     const { activeUserInfo, features } = useAppSelector((state: RootState) => state.app);
     const isAuthenticated = useIsAuthenticated();
-    const defaultUserInfoStatusText = 'Hang tight while we fetch your information...';
-    const [userInfoStatusText, setUserInfoStatusText] = React.useState<string>(defaultUserInfoStatusText);
 
     const chat = useChat();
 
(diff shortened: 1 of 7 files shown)


glahaye (approved):

webapp/src/App.tsx
```diff
@@ -143,10 +147,16 @@ const App: FC = () => {
                         <BackendProbe
                             uri={process.env.REACT_APP_BACKEND_URI as string}
                             onBackendFound={() => {
-                                setAppState(AppState.LoadingChats);
+                                setAppState(AppState.SettingUserInfo);
                             }}
                         />
                     )}
+                    {appState === AppState.SettingUserInfo &&
+                        (userInfoStatusText === defaultUserInfoStatusText ? (
```
TaoChenOSU, line 155: nit: is it a good practice to use string comparison rather than a flag to decide a state of the app?

TaoChenOSU: For example, we can have a flag `loadingUserInfoStatus`, with `true` being loading and `false` being failed. Then we can just pass the texts to `Loading` or `Error` instead of using `userInfoStatusText` in both places.

TaoChenOSU (dismissed):

dehoward (approved):

webapp/src/App.tsx
teresaqhoang: You're so right. Since we're transitioning to another view altogether for the error state, I added another property in the AppState enum called ErrorLoadingUserInfo and that's what we transition to on error. This will help force the user to sign-out and sign back in as well.

Thanks for the callout!

TaoChenOSU: This is even better! Thanks!

TaoChenOSU (approved):

PR #149, opened 2023-08-10
Adding option to view full stepwise context after every run

teresaqhoang:
### Motivation and Context

This PR adds an expandable/collapsible section to every stepwise planner result in the prompt dialog to show more details about the thought process.

Details for other semantic dependencies will be added in a follow-up PR: https://github.com/microsoft/chat-copilot/issues/150

### Description

Screenshot 2023-08-11 at 10 34 40 AM

### Contribution Checklist


webapi/Models/Response/BotResponsePrompt.cs
```diff
@@ -9,6 +9,29 @@ namespace CopilotChat.WebApi.Models.Response;
 /// </summary>
 public class BotResponsePrompt
 {
+    /// <summary>
+    /// Information about semantic dependencies of the prompt.
+    /// </summary>
+    public class SemanticDependency<T>
```
teresaqhoang, line 15: Context for other semantic dependencies will be added in a separate PR: https://github.com/microsoft/chat-copilot/issues/150

webapi/Models/Response/StepwiseThoughtProcess.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Steps taken execution stat.
+    /// </summary>
+    [JsonPropertyName("stepsTaken")]
+    public string StepsTaken { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Time taken to fulfil the goal.
+    /// </summary>
+    [JsonPropertyName("timeTaken")]
+    public string TimeTaken { get; set; } = string.Empty;
```
TaoChenOSU, line 22: nit: specify the unit in the comment since this is a string

webapi/Models/Response/BotResponsePrompt.cs
alliscode, line 15: Can we pull this out into its own file?

TaoChenOSU (dismissed):

webapi/Models/Response/BotResponsePrompt.cs
```diff
@@ -37,7 +60,7 @@ public class BotResponsePrompt
     /// Relevant additional knowledge extracted using a planner.
     /// </summary>
     [JsonPropertyName("externalInformation")]
-    public string ExternalInformation { get; set; } = string.Empty;
+    public SemanticDependency<StepwiseThoughtProcess> ExternalInformation { get; set; }
```
alliscode, line 63: Is BotResponsePrompt currently used for other context types? Will this break them in the time between checking this in and adding the others?

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -347,6 +348,13 @@ private async Task<ChatMessage> GetChatResponseAsync(string chatId, string userI
         var planResult = await this.AcquireExternalInformationAsync(chatContext, userIntent, externalInformationTokenLimit);
         chatContext.ThrowIfFailed();
 
+        // Extract additional details about planner execution in chat context
+        // TODO: [Issue #150, sk#2106] Accommodate different planner contexts once core team finishes work to return prompt and token usage.
+        var plannerDetails = new SemanticDependency<StepwiseThoughtProcess>(
+                planResult,
+                this._externalInformationSkill.StepwiseThoughtProcess
+            );
```
alliscode, line 356: Does this work when configured for other planner types?

alliscode (changes requested):

dehoward (approved):

webapi/Models/Response/BotResponsePrompt.cs
teresaqhoang: No, it's only used for the prompt view in the webapp and I've made that backwards compatible with string types too (since we were just returning the raw rendered string to show in the prompt dialog before)

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: No, we don't have the ability to get any more context from other planners, but the Core team has a work item to return more context including token usage, so I'll update this when that's done. There's a TODO to track that work.

Until then, context will only exist for Stepwise planner and the prompt dialog will have the same behavior it did before for other planners, which is displaying the proposed plan json.

webapi/Models/Response/BotResponsePrompt.cs
teresaqhoang: Done

webapi/Models/Response/StepwiseThoughtProcess.cs
teresaqhoang: Added format definition in comment

alliscode (approved):

TaoChenOSU (approved):

Issue #141, opened 2023-08-09
Update webapi routes


gitri-ms:
- Change routes to better align with Microsoft's API guidelines, per the suggestions in https://github.com/microsoft/semantic-kernel/pull/1090. This will be a breaking change.
- Routes that access a particular chat session should include the chatId in the route so that we can leverage the "RequireChatParticipant" authorization policy.

PR #152, opened 2023-08-11
Semantic Memory Integration into Copilot Chat

crickman:
### Motivation and Context

Preview of early break-through/poc 

### Description

- Exising API unchanged
- New route for import and status
- Ask/Search TBD

### Contribution Checklist


.github/workflows/copilot-build-backend.yml
```diff
@@ -24,6 +24,8 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
+    environment: feature-semantic-memory
```
dehoward, line 27: remove?

webapi/Controllers/ChatMemoryController.cs
```diff
@@ -84,24 +85,30 @@
...
-                memories.Add(memory.Metadata.Text);
+                memories.Add(memory.Text);
             }
         }
         catch (SKException connectorException)
         {
             // A store exception might be thrown if the collection does not exist, depending on the memory store connector.
-            var sanitizedMemoryCollectionName = memoryCollectionName.Replace(Environment.NewLine, string.Empty, StringComparison.Ordinal);
-            this._logger.LogError(connectorException, "Cannot search collection {0}", sanitizedMemoryCollectionName);
+            this._logger.LogError(connectorException, "Cannot search collection {0}", sanitizedMemoryName);
```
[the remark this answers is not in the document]
crickman: Ok ok...

.github/workflows/copilot-deploy-pipeline.yml
```diff
@@ -3,7 +3,7 @@ name: copilot-deploy-pipeline
 on:
   workflow_dispatch:
   push:
-    branches: ["main"]
+    branches: ["feature-semantic-memory"]
```
TaoChenOSU, line 6: This is temporary. Will change back to main before this PR is ready.

webapi/Controllers/DocumentController.cs
```diff
@@ -283,122 +376,122 @@ private async Task ValidateDocumentImportFormAsync(DocumentImportForm documentIm
...
-                            catch (Exception ex) when (!ex.IsCriticalException())
-                            {
-                                this._logger.LogError(ex, "Failed to analyze image {0} with Content Safety. ErrorCode: {{1}}", formFile.FileName, (ex as AIException)?.ErrorCode);
-                                throw new AggregateException($"Failed to analyze image {formFile.FileName} with Content Safety.", ex);
-                            }
+                var violations = new List<string>();
+                try
+                {
+                    // Call the content safety controller to analyze the image
+                    var imageAnalysisResponse = await this._contentSafetyService.ImageAnalysisAsync(formFile, default);
```
teresaqhoang, line 411: Does this work right now without the image type check? Or does it just throw an exception on any non-image file?

webapi/Extensions/ServiceExtensions.cs
```diff
@@ -123,30 +123,31 @@ internal static IServiceCollection AddCorsPolicy(this IServiceCollection service
     /// <exception cref="InvalidOperationException"></exception>
     public static IServiceCollection AddPersistentOcrSupport(this IServiceCollection services)
```
teresaqhoang, line 124: Why is OCR support still needed if the memory service handles all image processing?

webapi/Skills/ChatSkills/SemanticMemorySkill.cs
```diff
@@ -64,53 +65,118 @@ public class SemanticChatMemorySkill
...
-                this._logger.LogError(connectorException, "Cannot search collection {0}", memoryCollectionName);
+                if (!memoryMap.TryGetValue(this._promptOptions.DocumentMemoryName, out var memories))
+                {
+                    return;
+                }
+
+                foreach ((var memory, var citation) in memories)
+                {
+                    var memoryText = $"[{citation.SourceName}] {memory}\n";
+                    builderMemory.Append(memoryText);
```
teresaqhoang, line 111: There's a manual check in PromptDialog.tsx that checks for document memories from the preamble concatenated before -- is there a way to know document memories from semantic memory now?      

maybe add a Preamble here or separate the two?          

This is the check `!(value as string).includes('User has also shared some document snippets:')`

teresaqhoang (commented):
Currently, during deleteChat, memories are deleted by collection, which was formally prefixed with the chat Id. Moving forward, do we have to iterate through all memories and check the chat id tag?

webapi/Controllers/DocumentController.cs
```diff
@@ -125,12 +86,82 @@ public bool ContentSafetyStatus()
...
+                    };
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Service API for importing a document.
+    /// </summary>
+    [Route("document/import")]
```
gitri-ms, line 159: We should really think through these route changes to make sure they align with Microsoft's REST API standards. Route changes are breaking changes, so we need to make sure we do them right and do them once. We also need to provide mitigation steps for our customers to minimize downtime.

There are additional route changes we are planning to make (see issue #141) so it might be good to do them all together, so we don't break customers multiple times. Should discuss with @alliscode

crickman:
> Currently, during deleteChat, memories are deleted by collection, which was formally prefixed with the chat Id. Moving forward, do we have to iterate through all memories and check the chat id tag?

We can scope query by tag.  Unfortunately, the specifics on the delete contract haven't been exposed through the semantic-memory interface yet.

webapi/Controllers/DocumentController.cs
crickman: Thanks for the call-out on #141.  In terms of customer impact, I'd expect that app and api would be updated together.  (I mean we regularly make changes where an app deployment alone would break w/o updating API.)

We certainly need to confirm our thoughts around this...totally agree.

webapi/Skills/ChatSkills/SemanticMemorySkill.cs
crickman: Thanks for this call-out.  Right...I was wondering about the document memory separation...in this change its mixed by relevance...i think.  Agree this is an unnecessary change and will address.

webapi/Extensions/ServiceExtensions.cs
crickman: Its not, I just haven't ripped it all out yet since it wasn't clear how controversial moving OCR to semantic-memory might be.

https://github.com/microsoft/semantic-memory/pull/29

I hope to get this integrated today or Monday.

webapi/Controllers/DocumentController.cs
crickman: Yeah...I think it'll throw.  Yes...its a hacky merge as I tried to expose a contract we could use via semantic-memory:

https://github.com/microsoft/semantic-memory/pull/29

webapi/Controllers/DocumentController.cs
```diff
@@ -125,12 +86,82 @@ public bool ContentSafetyStatus()
...
+                }
+                else
+                {
+                    yield return new StatusResult(targetCollectionName, documentReference)
+                    {
+                        IsStarted = true,
+                        IsCompleted = status.Completed,
+                        LastUpdate = status.LastUpdate,
+                        Keys = Array.Empty<string>(), // $$$ DCR MEMORY | DCR CHAT
+                        Tokens = 0, // $$$ DCR MEMORY | DCR CHAT
```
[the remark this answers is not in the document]
crickman: Agree we need to solve the token-count thing.  Devis removed it from the semantic-memory model and I let it slip to the back-burner.

webapi/Controllers/DocumentController.cs
```diff
@@ -115,7 +76,7 @@ private enum SupportedFileType
     /// </summary>
     /// <returns></returns>
     [HttpGet]
-    [Route("contentSafety/status")]
+    [Route("document/safetystatus")]
```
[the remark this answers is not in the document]
crickman: I agree with your analysis...I wasn't satisfied with the route either.  Honestly I wonder if just having a stand-alone controller might make sense?  If you think so, what would that route be? (I mean, its not document specific).

memorypipeline/appsettings.json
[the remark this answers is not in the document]
crickman: Good call-out.  We've been merging/copying through many breaking changes from the semantic-memory config.  I do think its become more stable of the last week.  Enhancing the comments is a great point.

memorypipeline/appsettings.json
```diff
@@ -0,0 +1,132 @@
...
+      // Whether to run the asynchronous pipeline handlers
+      // Use these booleans to deploy the web service and the handlers on same/different VMs
+      "RunHandlers": true,
+      // Whether to expose OpenAPI swagger UI
+      "OpenApiEnabled": false
+    },
+    // "AzureBlobs" or "FileSystemContentStorage"
+    "ContentStorageType": "FileSystemContentStorage",
+    // Data ingestion pipelines configuration.
+    "DataIngestion": {
```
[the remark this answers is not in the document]
crickman: @TaoChenOSU is working on CC docs and I think we can also drive people to the semantic-memory repo/documentation as it evolves.

webapi/Controllers/DocumentController.cs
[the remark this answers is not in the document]
gitri-ms: Re: standalone controller, think about how we eventually might want to componentize the webapi to only add the controllers for features you enable. If it makes sense to enable content safety without doccuments (or vice versa), then they should be in separate controllers.

memorypipeline/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.SemanticMemory.Core.Diagnostics;
+
+var app = WebApplication.CreateBuilder().AddMemoryServices().Build();
+
+// ********************************************************
+// ************** START ***********************************
+// ********************************************************
+
+app.Logger.LogInformation(
+    "Starting Copilot Chat Memory pipeline service, .NET Env: {0}, Log Level: {1}",
```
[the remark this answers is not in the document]
crickman: (I just learned about the suggested changes feature - thank you!)

kbeaugrand:
Hi,

One question regarding this PR. By moving to semantic-memory, you remove the possibility to add new connectors directly to chat-copilot right?

We would only be able to use already implemented connectors in semantic-memory, if we want to leverage on other database, third-party connectors might be difficult to add in that case...

crickman:
> Hi,
> 
> One question regarding this PR. By moving to semantic-memory, you remove the possibility to add new connectors directly to chat-copilot right?
> 
> We would only be able to use already implemented connectors in semantic-memory, if we want to leverage on other database, third-party connectors might be difficult to add in that case...

This is correct.  In fact, due to a lack in parity we are temporarilly regressing support for Postgres and Chroma connectors.  I expect the memory-service to quickly match parity with what is currently supported in sk.  As the memory-service evolves to support more complex scenarios, then the value proposition may be quite a bit higher than the connector model.

But, yes, utimately chat-copilot will no longer make use of the `ISemanticTextMemory `interrace.

alliscode (approved):

alliscode (approved):

PR #165, opened 2023-08-14
Add version number information in backend, frontend and build process

glahaye:
### Motivation and Context
We want to make it easier to know which version we are dealing with / debugging

### Description

- Add version number information in backend, frontend and build process
- Simplify ServiceOptionsResponse and make it more generic
- Remove unused deployment files

image

image

### Contribution Checklist


webapi/Models/Response/ServiceOptionsResponse.cs
```diff
@@ -1,34 +1,18 @@
...
+/// Response given by <seealso cref="ServiceOptionsController"/>
 /// </summary>
-public class MemoryStoreOptionResponse
+public class ServiceOptionsResponse
 {
     /// <summary>
-    /// All the available memory store types.
-    /// </summary>
-    [JsonPropertyName("types")]
-    public IEnumerable<string> Types { get; set; } = Enumerable.Empty<string>();
```
alliscode, line 27: Isn't this needed to populate the list of supported options in the front-end?

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
```diff
@@ -93,6 +93,19 @@ export const SettingsDialog: React.FC<ISettingsDialogProps> = ({ open, closeDial
...
+                            <Divider />
+                            <AccordionItem value="about">
+                                <AccordionHeader expandIconPosition="end">
+                                    <h3>About</h3>
+                                </AccordionHeader>
+                                <AccordionPanel>
+                                    <Body1 color={tokens.colorNeutralForeground3}>
+                                        Frontend version: {process.env.REACT_APP_SK_VERSION}<br/>
+                                        {process.env.REACT_APP_SK_BUILD_INFO}<br/>
+                                        Backend version: {serviceOptions.values.version}
```
alliscode, line 105: What do these default to if built locally?

alliscode (changes requested):

.github/workflows/typos.yaml
```diff
@@ -1,29 +0,0 @@
-# Check pull requests for typos.
```
TaoChenOSU, line 1: Why are we removing this workflow?

.github/workflows/copilot-build-frontend.yml
```diff
@@ -1,35 +0,0 @@
-name: copilot-build-frontend
```
TaoChenOSU, line 1: Why are we removing this workflow?

webapi/Controllers/ServiceOptionsController.cs
```diff
@@ -34,20 +35,24 @@ public class ServiceOptionsController : ControllerBase
     /// <summary>
     /// Return the memory store type that is configured.
     /// </summary>
-    [Route("serviceOptions")]
+    [Route("ServiceOptions")]
```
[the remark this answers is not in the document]
glahaye: I'll revert that.

webapi/Models/Response/ServiceOptionsResponse.cs
```diff
@@ -1,34 +1,18 @@
...
-
-    /// <summary>
-    /// The selected memory store type.
+    /// Dictionary of values to return to caller.
     /// </summary>
-    [JsonPropertyName("selectedType")]
-    public string SelectedType { get; set; } = string.Empty;
+    [JsonPropertyName("values")]
+    public Dictionary<string, string> Values { get; set; } = new();
```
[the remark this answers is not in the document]
glahaye: Bringing back type contract.

webapi/Models/Response/ServiceOptionsResponse.cs
glahaye: It would be. But why do we even want to know all the other types of possible memory backings? The list keeps expanding and the UI will get more cluttered as we go while we only support using one memory type at a time.  I'm thinking just display what we are using and forget what we are not using.

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
glahaye: Blank for the first two and whatever the backend gives us (1.0.0 if nothing was specified while building the backend) for the last line.

I could put something like "Local build" for the frontend version if no version is provided.

The second line just doesn't appear if nothing is given.

.github/workflows/copilot-build-frontend.yml
glahaye: It actually isn't / shouldn't be used.

webapi/Models/Response/ServiceOptionsResponse.cs
alliscode: I understand your point, however chat-copilot is useful as a way to demonstrate the capabilities of semantic kernel. One thing we are focusing on is working with other platforms and this is a great way to demonstrate this. I agree that as the list grows, we will need a different way to show this off.

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
alliscode: Ok that's reasonable. Maybe defaulting to "Not provided" or something along those lines would make sense.

glahaye: Putting in "-" if nothing is specified

webapi/Models/Response/ServiceOptionsResponse.cs
glahaye: Reverting this

webapi/Models/Response/ServiceOptionsResponse.cs
```diff
@@ -1,34 +1,18 @@
...
 namespace CopilotChat.WebApi.Models.Response;
 
-public class ServiceOptionsResponse
-{
-    /// <summary>
-    /// Configured memory store.
-    /// </summary>
-    [JsonPropertyName("memoryStore")]
-    public MemoryStoreOptionResponse MemoryStore { get; set; } = new MemoryStoreOptionResponse();
-}
```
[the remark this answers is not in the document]
glahaye: Done

.github/workflows/typos.yaml
teresaqhoang: Isn't it used in the PR pipelines? Or is there another spell check that's configured there?

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
```diff
@@ -93,6 +93,19 @@ export const SettingsDialog: React.FC<ISettingsDialogProps> = ({ open, closeDial
...
+                                    <h3>About</h3>
+                                </AccordionHeader>
+                                <AccordionPanel>
+                                    <Body1 color={tokens.colorNeutralForeground3}>
+                                        Frontend version: {process.env.REACT_APP_SK_VERSION ?? "-"}<br/>
+                                        {process.env.REACT_APP_SK_BUILD_INFO}<br/>
+                                        Backend version: {serviceOptions.version}
+                                    </Body1>
+                                </AccordionPanel>
+                            </AccordionItem>
```
teresaqhoang, line 108: nit: add a divider after so there's separation with the footer when scrolled all the way down
```suggestion
                            </AccordionItem>
                            <Divider />
```
Screenshot 2023-08-15 at 10 06 31 AM

.github/workflows/typos.yaml
glahaye: If it's used, I can't find where. Maybe we should have something like that, though

scripts/deploy/deploy-webapp.ps1
```diff
@@ -27,7 +27,15 @@ param(
...
     # Authority for client applications that are not configured as multi-tenant.
-    $Authority="https://login.microsoftonline.com/common"
+    $Authority="https://login.microsoftonline.com/common",
+
+    [string]
+    # Version to display in UI.
+    $Version = "",
+
+    [string]
+    # Additional information given in version info.
```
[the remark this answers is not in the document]
glahaye: Added

.github/workflows/copilot-build-frontend.yml
dehoward: this workflow is and should be used. it is triggered by Github in PRs when any changes are made to the `webapp/` folder and is responsible for ensuring that the build/format is passing:
image

.github/workflows/typos.yaml
dehoward: it's triggered by Github for all PRs:
image

.github/workflows/copilot-build-frontend.yml
glahaye: @dehoward You are totally right!  For some reason, when I removed it, I just saw it as a dup of the frontend deployment workflow and didn't think of the PR workflow.  Restoring this file.

.github/workflows/copilot-deploy-frontend.yml
```diff
@@ -9,6 +9,9 @@ on:
       DEPLOYMENT_NAME:
         required: true
         type: string
+      MAJOR_MINOR:
```
dehoward, line 12: why not just `VERSION`?

scripts/deploy/package-webapi.ps1
```diff
@@ -37,7 +45,7 @@ if (!(Test-Path $publishOutputDirectory)) {
 }
 
 Write-Host "Build configuration: $BuildConfiguration"
-dotnet publish "$PSScriptRoot/../../webapi/CopilotChatWebApi.csproj" --configuration $BuildConfiguration --framework $DotNetFramework --runtime $TargetRuntime --self-contained --output "$publishOutputDirectory"
+dotnet publish "$PSScriptRoot/../../webapi/CopilotChatWebApi.csproj" --configuration $BuildConfiguration --framework $DotNetFramework --runtime $TargetRuntime --self-contained --output "$publishOutputDirectory" /p:AssemblyVersion=$Version /p:FileVersion=$Version /p:InformationalVersion=$InformationalVersion
```
dehoward, line 48: why do we need them both if they hold the same value?

.github/workflows/typos.yaml
glahaye: @dehoward Thanks! I didn't think of PR's for some reason... Reverting this deletion.

.github/workflows/copilot-deploy-frontend.yml
glahaye: Sure.

scripts/deploy/deploy-webapp.ps1
```diff
@@ -27,7 +27,15 @@ param(
...
-    $Authority="https://login.microsoftonline.com/common"
+    $Authority="https://login.microsoftonline.com/common",
+
+    [string]
+    # Version to display in UI.
+    $Version = "",
+
+    [string]
+    # Additional information given in version info. (Ex: commit SHA)
+    $InformationalVersion = ""
```
dehoward, line 38: can we make the name of this new arg more descriptive? (in the other files as well)
```suggestion
    $CommtiSha = ""
```

scripts/deploy/deploy-webapp.sh
```diff
@@ -15,6 +15,8 @@ usage() {
     echo "  -d, --deployment-name DEPLOYMENT_NAME  Name of the deployment from a 'deploy-azure.sh' deployment (mandatory)"
     echo "  -a, --application-id APPLICATION_ID    Client application ID (mandatory)"
     echo "  -au, --authority                       Authority to use for client applications that are not configured as multi-tenant. Defaults to (https://login.microsoftonline.com/common) if not specified."
+    echo "  -v  --version VERSION                  Version to set files to (default: 1.0.0)"
```
dehoward, line 18: nit:
```suggestion
    echo "  -v  --version VERSION                  Version of deployment (default: 1.0.0)"
```

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
```diff
@@ -93,6 +93,20 @@ export const SettingsDialog: React.FC<ISettingsDialogProps> = ({ open, closeDial
...
+                            <Divider />
+                            <AccordionItem value="about">
+                                <AccordionHeader expandIconPosition="end">
+                                    <h3>About</h3>
+                                </AccordionHeader>
+                                <AccordionPanel>
+                                    <Body1 color={tokens.colorNeutralForeground3}>
+                                        Frontend version: {process.env.REACT_APP_SK_VERSION ?? "-"}<br/>
+                                        {process.env.REACT_APP_SK_BUILD_INFO}<br/>
+                                        Backend version: {serviceOptions.version}
```
dehoward, line 105: nit: can we move this above the commit SHA so that both versions are next to each other?

dehoward (changes requested):

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
```diff
@@ -93,6 +93,20 @@ export const SettingsDialog: React.FC<ISettingsDialogProps> = ({ open, closeDial
...
                                 </AccordionPanel>
                             </AccordionItem>
+                            <Divider />
+                            <AccordionItem value="about">
+                                <AccordionHeader expandIconPosition="end">
+                                    <h3>About</h3>
+                                </AccordionHeader>
+                                <AccordionPanel>
+                                    <Body1 color={tokens.colorNeutralForeground3}>
+                                        Frontend version: {process.env.REACT_APP_SK_VERSION ?? "-"}<br/>
```
dehoward, line 103: instead of defining a new env variable for the version of the UX, I think we can pull this from the `version` field in `package.json`.

webapi/Controllers/ServiceOptionsController.cs
```diff
@@ -39,15 +41,24 @@ public class ServiceOptionsController : ControllerBase
...
+            Version = GetAssemblyFileVersion()
+        };
+
+        return this.Ok(response);
+    }
+
+    private static string GetAssemblyFileVersion()
+    {
+        Assembly assembly = Assembly.GetExecutingAssembly();
+        FileVersionInfo fileVersion = FileVersionInfo.GetVersionInfo(assembly.Location);
```
dehoward, line 60: does this refer to a different type of version that what is passed in during deployment?

scripts/deploy/package-webapi.ps1
glahaye: If we don't specify either, it will default to 1.0.0 and just be confusing:
https://learn.microsoft.com/en-us/troubleshoot/developer/visualstudio/general/assembly-version-assembly-file-version

webapi/Controllers/ServiceOptionsController.cs
glahaye: No.  This extracts the version that was set (or not) during build / deployment

glahaye:
> Is the version bumped automatically? Or is this something we have to do manually? If the latter, we should specify this in a contributing doc somewhere
> 
> should also start a version changelog in this PR

For proper Releases and change logs, we can use functionality baked in GitHub already. This PR is but a step for that to happen. Stay tuned.

scripts/deploy/deploy-webapp.ps1
glahaye: Well, anything arbitrary string can be passed here. In the deployment script, I just chose to pass the commit SHA for now because I thought it made sense.

scripts/deploy/deploy-webapp.sh
glahaye: Changed it to match what's in the ps1 file: Version to display in UI

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
glahaye: Done

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
glahaye: Intriguing idea. Are you proposing to do a second commit for every merge? Or intercepting merge requests and changing the version number then?

scripts/deploy/deploy-webapp.ps1
dehoward: yeah, that's fair. so maybe `VersionInfo`?

dehoward:
> > Is the version bumped automatically? Or is this something we have to do manually? If the latter, we should specify this in a contributing doc somewhere
> > should also start a version changelog in this PR
> 
> For proper Releases and change logs, we can use functionality baked in GitHub already. This PR is but a step for that to happen. Stay tuned.

so with that said, we will have automatic updating of the major/minors in the future?
I have concerns about having to manually update these fields, but they are alleviated if Github will does it for us 😄

webapp/src/components/header/settings-dialog/SettingsDialog.tsx
dehoward: my initial thought was the latter, to move the onus of keeping up with the version away from the developer. 
but I see that the sole purpose of this variable is to keep track of what's currently deployed, and in that case we shouldn't be touching any of the actual files.

resolving 😄

scripts/deploy/deploy-webapp.sh
```diff
@@ -15,6 +15,8 @@ usage() {
     echo "  -d, --deployment-name DEPLOYMENT_NAME  Name of the deployment from a 'deploy-azure.sh' deployment (mandatory)"
     echo "  -a, --application-id APPLICATION_ID    Client application ID (mandatory)"
     echo "  -au, --authority                       Authority to use for client applications that are not configured as multi-tenant. Defaults to (https://login.microsoftonline.com/common) if not specified."
+    echo "  -v  --version VERSION                  Version to display in UI (default: 1.0.0)"
+    echo "  -i  --info INFO                        Additional info to put in version details"
```
dehoward, line 19: same here `versionInfo`

scripts/deploy/deploy-webapp.ps1
glahaye: Deal! :)

scripts/deploy/deploy-webapp.sh
glahaye: Done

alliscode (approved):

dehoward (approved):

PR #178, opened 2023-08-15
Tuned: Document chunking parameters and debug output (customer input)

crickman:
### Motivation and Context

Responding to customer feedback on:

1. Document memory UX
1. Debug output

### Description

- When debugger is attached locally, appliication insight tracing is not written to the debug output.
- Larger document chunks with overlap provide a more coherent view of memories.  1000 tokens just isn't that much.

**Current:**
image
**Azure AI Playground:**
image
**Tuned:**
image

### Contribution Checklist

### DIFF
--- a/webapi/appsettings.json
+++ b/webapi/appsettings.json
@@ -161,8 +161,8 @@
   "DocumentMemory": {
     "GlobalDocumentCollectionName": "global-documents",
     "ChatDocumentCollectionNamePrefix": "chat-documents-",
-    "DocumentLineSplitMaxTokens": 100,
-    "DocumentParagraphSplitMaxLines": 1000,
+    "DocumentLineSplitMaxTokens": 64,
+    "DocumentParagraphSplitMaxLines": 512,
     "FileSizeLimit": 4000000,
     "FileCountLimit": 10
   },
--- a/webapi/appsettings.json
+++ b/webapi/appsettings.json
@@ -246,8 +246,5 @@
   // - Set "APPLICATIONINSIGHTS_CONNECTION_STRING" using dotnet's user secrets (see above)
   //     (i.e. dotnet user-secrets set "APPLICATIONINSIGHTS_CONNECTION_STRING" "MY_APPINS_CONNSTRING")
   //
-  "ApplicationInsights": {
-    "EnableDebugLogger": false
-  },
   "APPLICATIONINSIGHTS_CONNECTION_STRING": null
 }
--- a/webapi/appsettings.json
+++ b/webapi/appsettings.json
@@ -246,8 +246,8 @@
   // - Set "APPLICATIONINSIGHTS_CONNECTION_STRING" using dotnet's user secrets (see above)
   //     (i.e. dotnet user-secrets set "APPLICATIONINSIGHTS_CONNECTION_STRING" "MY_APPINS_CONNSTRING")
   //
-  //"ApplicationInsights": {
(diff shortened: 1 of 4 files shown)


glahaye (approved):

teresaqhoang (approved):

webapi/appsettings.json
```diff
@@ -161,8 +161,8 @@
   "DocumentMemory": {
     "GlobalDocumentCollectionName": "global-documents",
     "ChatDocumentCollectionNamePrefix": "chat-documents-",
-    "DocumentLineSplitMaxTokens": 30,
-    "DocumentParagraphSplitMaxLines": 100,
+    "DocumentLineSplitMaxTokens": 64,
+    "DocumentParagraphSplitMaxLines": 512,
```
alliscode, line 165: I think it's reasonable to change the defaults as you've done here. This of course depends a lot on the document that your feeding in so it would also help to add some more documentation, maybe just in the comments in this file, about the fact that these properties can be used to tune things.

I also think it could be useful to expose these parameters as part of the document import API so that they could change per document. We won't do that work now but we could create an item on the backlog and fit it in where it makes sense.

alliscode (approved):

webapi/appsettings.json
```diff
@@ -161,8 +161,8 @@
   "DocumentMemory": {
     "GlobalDocumentCollectionName": "global-documents",
     "ChatDocumentCollectionNamePrefix": "chat-documents-",
-    "DocumentLineSplitMaxTokens": 30,
-    "DocumentParagraphSplitMaxLines": 100,
+    "DocumentLineSplitMaxTokens": 64,
```
TaoChenOSU, line 164: We actually have defaults and documented why we selected these numbers here: https://github.com/microsoft/chat-copilot/blob/main/webapi/Options/DocumentMemoryOptions.cs#L32. Need to adjust the comments on why we selected the new numbers.

TaoChenOSU: Above this Json block, we also have a link to the OpenAI explaining why we selected the token limits for chunking. Will need to rewrite that too.

crickman: Thank you for this call-out.  Consistency is key, esp from customer perspective.

crickman: Aside: (I've read the material)  Part of the issue is that we've conflated our chunk boundary with _paragraph_.  A paragraph may very well be average 100 tokens, but this doesn't necessarily equivalent to document chunking granularity...esp for structured documents data.  In short, I don't sree that the [Open AI article](https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them) isn't directly recommending document chunking strategy.

We could rename the setting from `DocumentParagraphSplitMaxLines  `to `DocumentSplitMaxLines` ...also the parameter is actually `maxTokensPerParagraph`...so actually maybe  `DocumentSplitMaxTokens` ?

crickman: Updated

webapi/appsettings.json
crickman: Agree.  I think might want to consider how to drive these requirements into the semantic-memory since chunking logic will reside there.  (Currently, there is no way to specify the token threshold for semantic-memory.)

https://github.com/microsoft/chat-copilot/issues/180

TaoChenOSU (approved):

alliscode (approved):

PR #186, opened 2023-08-16
Allow CSV files to be imported as well

crgarcia12:
### Description

Simple extension to allow CSV files. Those files are treated as normal text files (like TXT and MD).
I am happy to provide a link to a deployed version of the project if anyone is interested in testing it out :)

### Contribution Checklist

- [ X ] The code builds clean without any errors or warnings
- [ X ] The PR follows the [Contribution Guidelines](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/copilot-chat/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [ X ] All unit tests pass, and I have added new tests where possible
- [ X ] I didn't break anyone :smile:

### DIFF
--- a/webapp/src/components/FileUploader.tsx
+++ b/webapp/src/components/FileUploader.tsx
@@ -45,7 +45,7 @@ export const FileUploader: React.FC<FileUploaderProps> = forwardRef<HTMLInputEle
                 className={classes.root}
                 accept={acceptedExtensions?.join(',')}
                 onChange={onChange}
-                title="Upload a .pdf, .txt, .jpg, .png or .tiff file"
+                title="Upload a .pdf, .txt, .jpg, .png .tiff or .csv file"
             />
         );
     },
--- a/webapi/Controllers/DocumentImportController.cs
+++ b/webapi/Controllers/DocumentImportController.cs
@@ -66,7 +66,12 @@ public class DocumentImportController : ControllerBase
         /// <summary>
         /// .tif or .tiff
         /// </summary>
-        Tiff
+        Tiff,
+
+        /// <summary>
+        /// .csv
+        /// </summary>
+        Csv,
     };
 
     private readonly ILogger<DocumentImportController> _logger;
@@ -270,6 +275,7 @@ public class DocumentImportController : ControllerBase
             switch (fileType)
             {
                 case SupportedFileType.Txt:
+                case SupportedFileType.Csv:
                 case SupportedFileType.Pdf:
                 case SupportedFileType.Md:
                     break;
@@ -307,6 +313,7 @@ public class DocumentImportController : ControllerBase
         {
             case SupportedFileType.Txt:
             case SupportedFileType.Md:
+            case SupportedFileType.Csv:
                 documentContent = await this.ReadTxtFileAsync(formFile);
                 break;
             case SupportedFileType.Pdf:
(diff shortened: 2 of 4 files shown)


webapp/src/components/FileUploader.tsx
```diff
@@ -45,7 +45,7 @@ export const FileUploader: React.FC<FileUploaderProps> = forwardRef<HTMLInputEle
                 className={classes.root}
                 accept={acceptedExtensions?.join(',')}
                 onChange={onChange}
-                title="Upload a .pdf, .txt, .jpg, .png or .tiff file"
+                title="Upload a .pdf, .txt, .jpg, .png .tiff or .csv file"
```
teresaqhoang, line 48: nit: 
```suggestion
                title="Upload a .pdf, .txt, .jpg, .png, .tiff or .csv file"
```

teresaqhoang (dismissed):

webapp/src/components/FileUploader.tsx
crgarcia12: I've added the missing comma, and moved CSV next to the text-type of files

crickman (changes requested):
This file-type logic is about to be removed from the application in favor of an evolution in our memory architecture.  Not sure if it makes sense to add right at this moment.

crickman:
Curious how you decided to deal with chunking/splitting the csv?  (how do chunks beyond the first express column header meta-data, if present?)

glahaye (approved):

glahaye:
> This file-type logic is about to be removed from the application in favor of an evolution in our memory architecture. Not sure if it makes sense to add right at this moment.

It doesn't hurt to have it until then. Then, we can port this just like the other text-based types

glahaye:
> Curious how you decided to deal with chunking/splitting the csv? (how do chunks beyond the first express column header meta-data, if present?)

Humm... I guess it would just create embeddings for each line, or set of lines...  Ideally, each line / set of lines would also have the column headers to give the lines meaning.

Other than that, the best is probably to have one embedding per line???

crgarcia12:
> Curious how you decided to deal with chunking/splitting the csv? (how do chunks beyond the first express column header meta-data, if present?)

In our case files are not that big. 
We even chunked as big as possible, always keeping the headers (first row and column) and it worked. 
With GPT-3 you can even ask for things like "how many Mercedes Benz EQA with more than 120 horse power are in the file?"

crgarcia12 (commented):
-

dluc:
I don't think this fits into the current design, which builds memory on top of unstructured data, ie natural language. 
Even if technically a CSV file can be chunked and embeddings can be generated, these will rarely match a query, and most of the time the CSV data will be left out from RAG due to a low cosine similarity. In worst case scenario only part of the data will be included with unpredictable results, depending on the content. This is a common problem with structured data, and it gets worse if the file contains opaque values without a clear meaning out of context. The problem appears also when dealing with tables inside Word/PDF docs, though in such cases the text around the tables provides useful context that RAG can use.
Long term, the plan for structured data (tables, CSV, Excel, SQL) is to use solutions such as nl2sql, which is a very different implementation compared to chunking and embedding.
My 2c, I would not merge the PR without testing, and I would recommend designing how nl2sql will deal with this type of content.

crickman:
I agree this change presents additional work around chunking and validation.  Considering this within the scope of the [semantic-memory project](https://github.com/microsoft/semantic-memory) seems appropriate:

PR #193, opened 2023-08-16
Delete Chat

teresaqhoang:
### Motivation and Context

This PR enables chat deletion. For V1, any user is able to delete any chat they're a part of. This will delete the chat for all participants, including Chat Copilot, as well as all related resources (memories, messages, etc.).

### Description

When a chat is deleted,
- It will be immediately removed for the user who initiated the deletion.
Screenshot 2023-08-16 at 5 54 46 PM

- If the chat was a multi-user chat, the chat will be disabled in current session. This allows other participants to save any resources before the chat is wiped out. On refresh, chat is removed from list.
Screenshot 2023-08-16 at 5 54 35 PM
Screenshot 2023-08-16 at 5 54 58 PM

- If the chat was the last chat in the user's current session, a new chat will automatically be created to populate the chat list.

### Contribution Checklist


webapi/Controllers/ChatHistoryController.cs
```diff
@@ -240,4 +241,78 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
 
         return this.NotFound($"No chat session found for chat id '{chatId}'.");
     }
+
+    /// <summary>
+    /// Delete a chat session.
+    /// </summary>
+    /// <param name="requestParameters">Object that contains the parameters to delete the chat.</param>
+    [HttpPost]
+    [Route("chatSession/delete")]
```
alliscode, line 250: To align with REST best practices of resources and collections of resources, can we make this an HttpDelete and the route "chatSessions/{sessionId}

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -240,4 +241,78 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+            chatToDelete = await this._sessionRepository.FindByIdAsync(chatId);
+        }
+        catch (KeyNotFoundException)
+        {
+            return this.NotFound($"No chat session found for chat id '{chatId}'.");
+        }
+
+        // Delete message and broadcast update to all participants.
+        await this._sessionRepository.DeleteAsync(chatToDelete);
+        await messageRelayHubContext.Clients.Group(chatId).SendAsync(ChatDeletedClientCall, chatId, userId);
```
alliscode, line 277: It might be good to move this to the last step. The reason is that something could fail at some point during the execution of all these tasks and in that case, it would be good for the frontend to still have a reference to the chat so that the user can try the delete again. As it is now, if something goes wrong right after the session is deleted, like the server loses power for example, then the pointer to the session is lost and all the participants, messages, and sources get orphaned.

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -240,4 +241,78 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+        var cleanupTasks = participantsTasks.Concat(messageTasks).Concat(sourceTasks);
+        await Task.WhenAll(cleanupTasks);
+
+        // Iterate over the tasks and check their status and exception
+        foreach (var task in cleanupTasks)
+        {
+            if (task.IsFaulted && task.Exception != null)
+            {
+                this._logger.LogInformation("Failed to delete an entity of chat {0}: {1}", chatId, task.Exception.Message);
+            }
```
alliscode, line 313: We should return a 500 error in this case so that the user can retry the delete.

alliscode (changes requested):

webapi/Models/Request/DeleteChatRequest.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// Id of the user who initiated chat deletion.
+    /// </summary>
+    [JsonPropertyName("userId")]
+    public string? UserId { get; set; }
+
+    /// <summary>
+    /// Id of the chat to delete.
+    /// </summary>
+    [JsonPropertyName("chatId")]
+    public string? ChatId { get; set; }
```
alliscode, line 22: This can be moved to the route as I mentioned above.

alliscode (dismissed):

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -240,4 +241,78 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+    /// <summary>
+    /// Delete a chat session.
+    /// </summary>
+    /// <param name="requestParameters">Object that contains the parameters to delete the chat.</param>
+    [HttpPost]
+    [Route("chatSession/delete")]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> DeleteChatSessionAsync([FromServices] IHubContext<MessageRelayHub> messageRelayHubContext, [FromBody] DeleteChatRequest requestParameters)
```
glahaye, line 254: Once Gina's auth and security changes are in, we'll only need to provide the chat ID, I believe, as the user id will be retrievable from the auth...

Just an FYI for the future.

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: Good point, I frontloaded resource deletion + added retry option on webapp if any part fails.

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: Good call, I'll add a TODO here to handle that

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -240,4 +241,100 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+    }
+
+    /// <summary>
+    /// Deletes all associated resources (messages, memories, participants) associated with a chat session.
+    /// </summary>
+    /// <param name="sessionId">The chat id.</param>
+    [HttpDelete]
+    [Route("chatSession/{sessionId:guid}/resources")]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
```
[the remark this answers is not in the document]
teresaqhoang: Oh yes, I was calling this from the webapp but since I frontloaded the resource deletion, we can retry chat deletion altogether on failure and this is no longer needed.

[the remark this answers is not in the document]
teresaqhoang: Removed

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: Returning a 500 now

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -240,4 +241,96 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+        if (userId == null)
+        {
+            return this.BadRequest("UserId cannot be null.");
+        }
+
+        ChatSession? chatToDelete = null;
+        try
+        {
+            // Make sure the chat session exists
+            chatToDelete = await this._sessionRepository.FindByIdAsync(chatId);
```
TaoChenOSU, line 269: We probably want to validate that the user is part of the chat before proceeding. Right now, the service will take the delete request from anyone even though they are not part of the chat.

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -240,4 +241,96 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+        await messageRelayHubContext.Clients.Group(chatId).SendAsync(ChatDeletedClientCall, chatId, userId);
+
+        return this.NoContent();
+    }
+
+    /// <summary>
+    /// Deletes all associated resources (messages, memories, participants) associated with a chat session.
+    /// </summary>
+    /// <param name="sessionId">The chat id.</param>
+    private async Task<IActionResult> DeleteChatResourcesAsync([FromServices] IHubContext<MessageRelayHub> messageRelayHubContext, Guid sessionId)
```
TaoChenOSU, line 294: I don't see this is removing memory.

webapp/src/assets/strings.ts
```diff
@@ -0,0 +1,8 @@
+export const COPY = {
+    STEPWISE_RESULT_NOT_FOUND_REGEX: /(Result not found, review _stepsTaken to see what happened\.)\s+(\[{.*}])/g,
+    CHAT_DELETED_MESSAGE: (chatName?: string) =>
+        `Chat ${
+            chatName ? `{${chatName}} ` : ''
+        }has been deleted by another user. Please save any resources you need and refresh the page.`,
```
TaoChenOSU, line 6: I wonder what resources they can save. The chat has been deleted at this point from the backend.

webapp/src/libs/hooks/useChat.ts
```diff
@@ -355,6 +359,32 @@ export const useChat = () => {
...
 
+    const deleteChat = async (chatId: string) => {
+        const friendlyChatName = getFriendlyChatName(conversations[chatId]);
+        await chatService
+            .deleteChatAsync(chatId, loggedInUser.id, await AuthHelper.getSKaaSAccessToken(instance, inProgress))
+            .then(() => {
+                dispatch(deleteConversation(chatId));
+
+                // If there is only one chat left, create a new chat
+                if (Object.keys(conversations).length <= 1) {
```
TaoChenOSU, line 370: This is more of question: will `conversations` get updated after the `dispatch` call?

webapi/Controllers/ChatHistoryController.cs
TaoChenOSU: Ahh, just saw your description. Resolving this one.

webapi/Controllers/ChatHistoryController.cs
crickman: Same question.

webapp/src/assets/strings.ts
teresaqhoang: If they want to save message content, screenshots, plan jsons, etc. 

I had it automatically delete chats for all users once the deletion was successful, and just having the chat removed when you're not the initiator was a bit jarring, so I wanted to give the chances to interact with it and refresh on their own accord.

webapp/src/libs/hooks/useChat.ts
teresaqhoang: yeah state will automatically get updated if any of its dependencies change, this is handled by redux

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,105 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+    /// <param name="sessionId">The chat id.</param>
+    [HttpDelete]
+    [Route("chatSession/{sessionId:guid}")]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> DeleteChatSessionAsync([FromServices] IHubContext<MessageRelayHub> messageRelayHubContext, [FromBody] DeleteChatRequest requestBody, Guid sessionId, CancellationToken cancellationToken)
+    {
+        var chatId = sessionId.ToString();
+        var userId = requestBody.UserId;
```
gitri-ms, line 285: Do not pass the userId in the request, instead get it from the AuthInfo object.

teresaqhoang: @gitri-ms I just pushed a change that uses AuthInfo - can you take another look?

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: Good catch! Removed memory collections tied to the chat

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,108 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+        [FromServices] IAuthInfo authInfo,
+        Guid sessionId,
+        CancellationToken cancellationToken)
+    {
+        var chatId = sessionId.ToString();
+
+        if (!(await this._participantRepository.IsUserInChatAsync(authInfo.UserId, chatId)))
+        {
+            return this.Forbid("User is unauthorized to delete this chat.");
+        }
```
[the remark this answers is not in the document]
teresaqhoang: Oh awesome! Added

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,108 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
 
         return this.NotFound($"No chat session found for chat id '{chatId}'.");
     }
+
+    /// <summary>
+    /// Delete a chat session.
+    /// </summary>
+    /// <param name="userId">Unique Id of user who initiated the request.</param>
```
gitri-ms, line 275: Remove

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,108 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+    /// <param name="sessionId">The chat id.</param>
+    [HttpDelete]
+    [Route("chatSession/{sessionId:guid}")]
```
gitri-ms, line 278: I feel like other places we have used 1chatId` instead of `sessionId`. Can we be consistent with our naming?

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,108 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> DeleteChatSessionAsync(
+        [FromServices] IHubContext<MessageRelayHub> messageRelayHubContext,
+        [FromServices] IAuthInfo authInfo,
```
gitri-ms, line 285: This was already passed into the constructor and stored as a member variable, so you don't need to pass it in here.

gitri-ms (changes requested):

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,105 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+    /// <summary>
+    /// Delete a chat session.
+    /// </summary>
+    /// <param name="userId">Unique Id of user who initiated the request.</param>
+    /// <param name="sessionId">The chat id.</param>
+    [HttpDelete]
+    [Route("chatSession/{sessionId:guid}")]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
```
glahaye, line 281: Looks like this call can also return 500...

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,103 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+        catch (KeyNotFoundException)
+        {
+            return this.NotFound($"No chat session found for chat id '{chatId}'.");
+        }
+
+        // Delete any resources associated with the chat session.
+        var deleteResourcesResult = await this.DeleteChatResourcesAsync(sessionId, cancellationToken) as StatusCodeResult;
+        if (deleteResourcesResult?.StatusCode != 204)
+        {
+            return this.StatusCode(500, $"Failed to delete resources for chat id '{chatId}'.");
```
glahaye, line 305: If this happens, you'll also want to log it, along with the status code

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,103 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+    private async Task<IActionResult> DeleteChatResourcesAsync(Guid sessionId, CancellationToken cancellationToken)
+    {
+        var chatId = sessionId.ToString();
+        var cleanupTasks = new List<Task>();
+
+        // Create and store the tasks for deleting all users tied to the chat.
+        var participants = await this._participantRepository.FindByChatIdAsync(chatId);
+        foreach (var participant in participants)
+        {
+            cleanupTasks.Add(this._participantRepository.DeleteAsync(participant));
```
glahaye, line 328: Just confirming...  These are the opaque GUID users, right?

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: We log individual resource failures in DeleteChatResources
image

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: What do you mean by opaque?

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: Renamed to chatId

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: It's the default user Id, if that's what you mean. There should only be one chat participant per chat session if the user is in Guest mode

image

image

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,4 +268,103 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
...
+        {
+            cleanupTasks.Add(this._memoryStore.DeleteCollectionAsync(collection, cancellationToken));
+        }
+
+        // Await all the tasks in parallel and handle the exceptions
+        await Task.WhenAll(cleanupTasks);
+        var failedTasks = false;
+
+        // Iterate over the tasks and check their status and exception
+        foreach (var task in cleanupTasks)
```
glahaye, line 357: If any of your tasks raises an exception, you will not reach this code.

See following for details:
https://www.davidguida.net/exception-handling-in-parallel-tasks/

webapp/src/Constants.ts
```diff
@@ -50,5 +50,4 @@ export const Constants = {
         MANIFEST_PATH: '/.well-known/ai-plugin.json',
     },
     KEYSTROKE_DEBOUNCE_TIME_MS: 250,
-    STEPWISE_RESULT_NOT_FOUND_REGEX: /(Result not found, review _stepsTaken to see what happened\.)\s+(\[{.*}])/g,
```
glahaye, line 53: Doesn't look related but I'm assuming that's fine?

teresaqhoang: Yeah I moved it to the strings file in a previous PR, this was the result of merging main

webapi/Controllers/ChatHistoryController.cs
teresaqhoang: Changed to handle exceptions in a try/catch block

webapi/Controllers/ChatHistoryController.cs
glahaye: OK. I'm good with that!

glahaye (dismissed):

webapp/src/assets/strings.ts
TaoChenOSU: I see. I think the resource is a bit misleading. It makes me think maybe I can save something to my account to download something to my local computer.

nit: change it to something like: ...has been deleted by another user. You can still view the most recent chat history at this point. The content will disappear once you refresh or close the app.

webapp/src/libs/hooks/useChat.ts
TaoChenOSU: If that's the case, why `<= 1`? Doesn't that mean it may still have 1 chat?

teresaqhoang: Oh sorry, I was wrong before. Conversations is not automatically updated; if I change the logic to `                if (Object.keys(conversations).length == 0) {` then it doesn't automatically create a new chat if the last one is updated. The method isn't updated, so it maintains the same execution thread so it's using the same conversations object.

webapp/src/assets/strings.ts
teresaqhoang: Changed!

webapp/src/libs/hooks/useChat.ts
TaoChenOSU: Got it!

TaoChenOSU (approved):

gitri-ms (approved):

PR #206, opened 2023-08-17
Build webapi package and deploy Azure infrastructure in parallel

glahaye:
### Motivation and Context
Faster deployments are always nicer.

### Description
Build webapi package and deploy Azure infrastructure in parallel instead of having infrastructure deployment wait for webapi package

### Contribution Checklist

### DIFF
--- a/.github/workflows/copilot-deploy-binaries.yml
+++ b/.github/workflows/copilot-deploy-binaries.yml
@@ -1,4 +1,4 @@
-name: copilot-deploy-environment
+name: copilot-deploy-binaries
 
 on:
   workflow_call:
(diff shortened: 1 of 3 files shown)


.github/workflows/copilot-deploy-binaries.yml
```diff
@@ -52,7 +40,6 @@ jobs:
       AZURE_SUBSCRIPTION_ID: ${{secrets.AZURE_SUBSCRIPTION_ID}}
 
   deploy-frontend:
-    needs: [deploy-infra]
     uses: ./.github/workflows/copilot-deploy-frontend.yml
     with:
       DEPLOYMENT_NAME: ${{needs.deploy-infra.outputs.deployment-id}}
```
alliscode, line 45: This won't work without having deploy-infra in the needs collection. `{{ needs.deploy-infra...`

.github/workflows/copilot-deploy-pipeline.yml
```diff
@@ -17,9 +17,20 @@ jobs:
   build:
     uses: ./.github/workflows/copilot-build-backend.yml
 
-  int:
-    needs: build
```
alliscode, line 21: I'm not sure it's a good idea to deploy the infra before we verify the code builds.

alliscode (changes requested):

.github/workflows/copilot-deploy-binaries.yml
alliscode: Currently the front-end and back-end deployments use the deployment Id to get some information about the resources so they need a reference to it. In theory you could pass this in as a parameter. However, I'm not sure we want to do this before the build has succeeded.

glahaye: Argh. Of course!  Whst did I tell you about GH's pipeline syntax...

Abandonning this PR

glahaye:
Abandoning this dubious change

PR #216, opened 2023-08-18
Add a dotnet format check workflow

TaoChenOSU:
### Motivation and Context

There is currently no C# format check workflow for PRs. There is a workflow step in copilot-build-backend.yml that checks the format for the webapi, but that happens after the build step.

There are two .Net projects in the repo (soon to be three). It's better to have a workflow that checks the format for the entire repo, instead of having multiple steps scattered around different workflows.

### Contribution Checklist

### DIFF
--- a/.github/workflows/dotnet-format.yml
+++ b/.github/workflows/dotnet-format.yml
@@ -59,7 +59,7 @@ jobs:
           fi
           csproj_files=($(printf "%s\n" "${csproj_files[@]}" | sort -u))
           echo "Found ${#csproj_files[@]} unique csproj/sln files: ${csproj_files[*]}"
-          echo "::set-output name=csproj_files::${csproj_files[*]}"
+          echo "csproj_files=${csproj_files[*]}" >> $GITHUB_OUTPUT
 
       - name: Setup .NET
         uses: actions/setup-dotnet@v1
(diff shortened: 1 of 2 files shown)


crickman (approved):

.github/workflows/dotnet-format.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: No C# files changed
+        id: no-csharp
+        if: github.event_name == 'pull_request' && steps.changed-files.outputs.added_modified == ''
+        run: echo "No C# files changed"
+
+      # This step will loop over the changed files and find the nearest .csproj file for each one, then store the unique csproj files in a variable
+      - name: Find csproj files
+        id: find-csproj
+        if: github.event_name != 'pull_request' || steps.changed-files.outputs.added_modified != '' || steps.changed-files.outcome == 'failure'
```
glahaye, line 40: I don't understand this condition...

We only run the following code if we DON'T have a PR or we have NO modified files OR we have a failure???

Is this correct?  Maybe add a comment?

TaoChenOSU: Good catch!

glahaye (approved):

Issue #238, opened 2023-08-22
CosmosDb partitioning strategy causing high cost/low perf, max quota/too many requests


dluc:
1. `chatsessions` table is partitioned by ID, but should be partitioned **by User** and have an index on the ID
2. `chatmessages` table is partitioned by ID, but should be partitioned **by Chat ID**

In the current setup, as these tables grow, CosmosDb takes more and more time to execute these operations:

1. Load the list of chats for the current user
2. Load the list of messages for the selected chat

Other than increasing **latency**, also the incurred **cost in RUs (and $)** grows, with the risk of hitting Max Quota when autoscale is not enabled.

crickman:
Thank you for reporting this.  I think we may also want to consider how to apply any fix in a manner that doesn't break existing chats.

crickman:
@dluc - Did you mean `chatparticipants` intead of `chatsessions` (no user in `chatsessions`):

> 1. `chatsessions` table is partitioned by ID, but should be partitioned **by User** and have an index on the ID

crickman:
Linked branch with core fix:

CosmosDB database and container are created during deployment (not by code).  The partitioning scheme is defined at this time in [scripts/deploy/main.bicep](https://github.com/microsoft/chat-copilot/blob/fix-cosmosdb-partitioning/scripts/deploy/main.bicep).

If no migration is required (as Devis stated), the containers can simply be deleted and redeployed.

The proper partition key will be utilized upon entity insert (no code change required).

Existing chat-copilot deployments will continue to work, but using the non-performant partitioning.

If migration pathway is required or some other type of schema-version detection, let's identify and iterate on those cases.

crickman:
Draft PR - https://github.com/microsoft/chat-copilot/pull/240

Following up w/ Tao tomorrow to explore deployment options to realize this change w/ minimal fuss.

PR #240, opened 2023-08-23
Update partitionKey path definitions for cosmosDatabase containers.

crickman:
### Motivation and Context
Customer issue reported in https://github.com/microsoft/chat-copilot/issues/238 related to partition scheme in cosmosdb containers.

### Description
Update partition-key path definition for `ChatMessage`, `ChatPartitipant`, and `MemorySource`.

Existing deployments will continue to work in their current mode.

CosmosDB containers must be removed and redeployed to realize the peformance update related to the partition scheme.

### Contribution Checklist


dluc:
is there any table that should be partitioned by user?

webapi/Storage/Repository.cs
```diff
@@ -42,22 +42,21 @@ public Task DeleteAsync(T entity)
...
-    public Task<T> FindByIdAsync(string id)
+    public Task<T> FindByIdAsync(string id, string? partition = null)
     {
-        return this.StorageContext.ReadAsync(id);
+        return this.StorageContext.ReadAsync(id, partition ?? id);
     }
 
     /// <inheritdoc/>
-    public async Task<bool> TryFindByIdAsync(string id, Action<T?> entity)
+    public async Task<bool> TryFindByIdAsync(string id, string? partition = null, Action<T?>? callback = null)
```
dluc, line 51: I think we need to pass the partition key, otherwise CosmosDb will have to scan all the partitions on every request

webapi/Controllers/ChatHistoryController.cs
```diff
@@ -260,7 +260,7 @@ public async Task<IActionResult> GetAllChatSessionsAsync(string userId)
     {
         this._logger.LogInformation("Get imported sources of chat session {0}", chatId);
 
-        if (await this._sessionRepository.TryFindByIdAsync(chatId.ToString(), v => _ = v))
+        if (await this._sessionRepository.TryFindByIdAsync(chatId.ToString()))
```
dluc, line 263: see also other comment: I believe we need to pass the PK, otherwise we're not addressing the perf problem. Either the client passes both values, or the "chat id" value should be a composite of PK+ID that's unpacked here

webapi/Models/Storage/ChatParticipant.cs
```diff
@@ -26,6 +27,12 @@ public class ChatParticipant : IStorageEntity
     /// </summary>
     public string ChatId { get; set; }
 
+    /// <summary>
+    /// The partition key for the source.
+    /// </summary>
+    [JsonIgnore]
+    public string Partition => this.ChatId;
```
TaoChenOSU, line 34: Should participants partitioned by user id? A user can be multiple participants in separate chats. We use participants to locate the chats a user belongs to.

crickman: All the models/entities are "chat-centric".  When querying this container entity, look-up occurs by chatid or chatid & userid.

webapi/Controllers/ChatHistoryController.cs
crickman: TryFindByIdAsync is primarily called for the ChatSession entity.  This entity doesn't have a partition key distinct from the primary key.  As such nearly all calls to this method end up looking like: `TryFindByIdAsync(chatId, chatId)` .. which just reads a little redundant.  In the underlying code, the pk value is coallecesed as the partition key when not provided:

   ```
 public async Task<bool> TryFindByIdAsync(string id, string? partition = null, Action<T?>? callback = null)
    {
        try
        {
            callback?.Invoke(await this.FindByIdAsync(id, partition ?? id));
            return true;
        }
        catch (Exception ex) when (ex is ArgumentOutOfRangeException || ex is KeyNotFoundException)
        {
            return false;
        }
    }

```

File: scripts/deploy/main.bicep
Comment (dluc) [inline scripts/deploy/main.bicep:1]: should the "chatsessions" table be partioned too, maybe by user?

**Comment by crickman** [comment]
> TryFindByIdAsync



> is there any table that should be partitioned by user?

No sir...this model is "chat-centric"

File: scripts/deploy/main.bicep [hunk 4]
```diff
@@ -637,7 +637,7 @@ resource messageContainer 'Microsoft.DocumentDB/databaseAccounts/sqlDatabases/co
       }
       partitionKey: {
         paths: [
-          '/id'
+          '/chatId'
```
dluc, line 640: need: what do you think about renaming "chatId" to "chatSessionId"? the current naming leads to some confusion, giving the impression that there's a "chats" table

scripts/deploy/main.bicep
crickman: I don't believe there is a user field in this entity...or any other partition candidate.  Am I missing something?

dluc:
Another perf detail that came up is the consistency level. The tables are setup for "strong consistency" which could be relaxed to reduce latency.

crickman:
Tested locally using https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/5b742c40-bc2b-4a4f-902f-ee9f644d8844/resourceGroups/crickman/providers/Microsoft.DocumentDB/databaseAccounts/crickman/overview

dluc (dismissed):

scripts/deploy/main.bicep
```diff
@@ -594,7 +594,7 @@ resource cosmosAccount 'Microsoft.DocumentDB/databaseAccounts@2023-04-15' = if (
   location: location
   kind: 'GlobalDocumentDB'
   properties: {
-    consistencyPolicy: { defaultConsistencyLevel: 'Session' }
+    consistencyPolicy: { defaultConsistencyLevel: 'Eventual' }
```
alliscode, line 597: I agree that we should move to eventual consistency... Changing from Strong to Eventual though can create some subtle bugs that are intermittent and depend on the timing of calls. Generally the client is build assuming eventual consistency and handles retries for scenarios where it matters such as a post to create a resource directly followed by a get for the same resource. I'm not sure if we have any cases like that currently.

I'm fine with checking this in if we've done a decent walkthrough of the app features and made sure we don't get any weirdness.

alliscode (approved):

scripts/deploy/main.bicep
crickman: Ok, back to `Session`

alliscode (approved):

PR #263, opened 2023-08-28
render chat messages as markdown

Rainson12:
### Description

When requesting `generate a table of fruits` or `generate some code to count from 1 to 200` the response will contain markdown: 
image

However Markdown can not be rendered in the frontend. This PR fixes it by adding a markdown rendering library and making use of it:
image

This also avoids using `dangerouslySetInnerHTML` attribute. The library `remark-gfm` is used because `react-markdown` does not support tables out of the box. 

This is how urls or links would be rendered:
image

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -27,7 +27,9 @@
     "microsoft-cognitiveservices-speech-sdk": "^1.31.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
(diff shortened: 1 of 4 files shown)


crickman (approved):
I like it.  Love the removal of `convertToAnchorTags`.

Do we want to run this through function testing?  @teresaqhoang ?

alliscode (approved):

PR #303, opened 2023-09-05
Fix formatting of Aad url in react env variables

jimmystridh:
### Motivation and Context

The powershell script does not include the protocol when generating the `REACT_APP_AAD_AUTHORITY` environment variable and the .env file default is based on this notion.

Without this fix, the resulting webapp/.env will look something like this:
```
> cat webapp/.env
REACT_APP_BACKEND_URI=https://localhost:40443/
REACT_APP_AUTH_TYPE=AzureAd
REACT_APP_AAD_AUTHORITY=https://https://login.microsoftonline.com/XXX
```

so the protocol will be duplicated into an invalid url.

### Description

The change removes the prefixing `https://` in the `configure.sh` script to match the behavior of the `Configure.ps1` script.

### Contribution Checklist


glahaye (approved):

gitri-ms (approved):

gitri-ms:
Thanks @jimmystridh for catching this!

PR #304, opened 2023-09-05
Adding Timeout Limit option for /chat requests

teresaqhoang:
### Motivation and Context

This PR adds a TimeoutLimitInS option in appsettings, which is used as the CancellationToken timeout if specified. The CancellationToken is used to monitor for cancellation requests.

### Description

This PR fixes a bug by passing the cancellation token to all async methods of ChatSkill. If a timeout limit is not specified in appsettings, a cancellation token source without a timeout is created and passed.

### Contribution Checklist

### DIFF
--- a/webapi/Controllers/ChatController.cs
+++ b/webapi/Controllers/ChatController.cs
@@ -8,6 +8,7 @@ using System.Net.Http;
 using System.Reflection;
 using System.Text.Json;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Threading.Tasks;
 using CopilotChat.WebApi.Auth;
 using CopilotChat.WebApi.Hubs;
@@ -131,7 +132,12 @@ public class ChatController : ControllerBase, IDisposable
         SKContext? result = null;
         try
         {
-            result = await kernel.RunAsync(contextVariables, function!);
+            // Define a timeout duration, e.g. 60 seconds
+            var timeout = TimeSpan.FromSeconds(60);
+
+            // Create a cancellation token source with the timeout
+            using var cts = new CancellationTokenSource(timeout);
+            result = await kernel.RunAsync(function!, contextVariables, cts.Token);
         }
         finally
         {
@@ -143,6 +149,13 @@ public class ChatController : ControllerBase, IDisposable
             if (result.LastException is AIException aiException && aiException.Detail is not null)
             {
                 return this.BadRequest(string.Concat(aiException.Message, " - Detail: " + aiException.Detail));
+            }
+
+            if (result.LastException?.InnerException is OperationCanceledException)
+            {
+                // Log the timeout and return a 408 response
+                this._logger.LogError("The chat operation timed out.");
+                return this.StatusCode(StatusCodes.Status408RequestTimeout);
             }
 
             return this.BadRequest(result.LastException!.Message);
--- a/webapi/Controllers/ChatController.cs
+++ b/webapi/Controllers/ChatController.cs
@@ -78,6 +78,7 @@ public class ChatController : ControllerBase, IDisposable
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status504GatewayTimeout)]
     public async Task<IActionResult> ChatAsync(
         [FromServices] IKernel kernel,
         [FromServices] IHubContext<MessageRelayHub> messageRelayHubContext,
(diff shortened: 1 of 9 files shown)


webapi/Controllers/ChatController.cs
```diff
@@ -145,6 +156,13 @@ public ChatController(ILogger<ChatController> logger, ITelemetryService telemetr
                 return this.BadRequest(string.Concat(aiException.Message, " - Detail: " + aiException.Detail));
             }
 
+            if (result.LastException is OperationCanceledException || result.LastException?.InnerException is OperationCanceledException)
+            {
+                // Log the timeout and return a 408 response
+                this._logger.LogError("The chat operation timed out.");
+                return this.StatusCode(StatusCodes.Status408RequestTimeout);
```
glahaye, line 163: You probably want 504 here, not 408.  408 means the client is the one giving the request too slowly and the server has had enough of your hogging a port.

504 means the server couldn't complete the request in time because one of its dependencies (e.g. AOAI) is taking too much time to return.

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -371,20 +380,24 @@ private async Task<ChatMessage> GetChatResponseAsync(string chatId, string userI
...
             // Save a new response to the chat history with the proposed plan content
             return await this.SaveNewResponseAsync(
-                JsonSerializer.Serialize<ProposedPlan>(proposedPlan), prompt, chatId, userId,
+                JsonSerializer.Serialize<ProposedPlan>(proposedPlan),
+                prompt,
+                chatId,
+                userId,
                 // TODO: [Issue #2106] Accommodate plan token usage differently
-                this.GetTokenUsagesAsync(chatContext)
+                this.GetTokenUsages(chatContext),
```
glahaye, line 388: Maybe the following is better:

```suggestion
                await this.GetTokenUsagesAsync(chatContext),
```

teresaqhoang: The method isn't asynchronous tho, this was a slip on my part when implementing it
image

webapp/src/libs/services/BaseService.ts
```diff
@@ -48,6 +48,11 @@ export class BaseService {
 
             if (!response.ok) {
                 const responseText = await response.text();
+
+                if (response.status === 408) {
```
glahaye, line 52: 408 to 504

webapi/Skills/ChatSkills/ChatSkill.cs
glahaye: Never mind, then!

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -464,11 +477,12 @@ private async Task<ChatMessage> GetChatResponseAsync(string chatId, string userI
     /// Helper function that creates the correct context variables to
     /// extract the audience from a conversation history.
     /// </summary>
-    private async Task<string> GetAudienceAsync(SKContext context)
+    /// <param name="cancellationToken">The cancellation token.</param>
+    private async Task<string> GetAudienceAsync(SKContext context, CancellationToken cancellationToken = default)
```
dehoward, line 481: do we need `= default` in all of the private functions? I see its there in `ChatAsync` but not in `GetChatResponseAsync` which seems to make sense

teresaqhoang: Really good point, it was originally specified to follow the same default conventions of SK. 

I removed the `= default` from the private methods. Read more about the differences, and since cancellation can be critical if the timeout is set, I want to pivot towards reliability and propagation over flexibility and compatibility. Thanks for the comment!

webapi/Controllers/ChatController.cs
teresaqhoang: Makes total sense. Changed! Thanks for calling this out!

webapi/Controllers/ChatController.cs
```diff
@@ -145,6 +156,13 @@ public ChatController(ILogger<ChatController> logger, ITelemetryService telemetr
                 return this.BadRequest(string.Concat(aiException.Message, " - Detail: " + aiException.Detail));
             }
 
+            if (result.LastException is OperationCanceledException || result.LastException?.InnerException is OperationCanceledException)
+            {
+                // Log the timeout and return a 504 response
+                this._logger.LogError("The chat operation timed out.");
+                return this.StatusCode(StatusCodes.Status504GatewayTimeout, "The chat operation timed out.");
```
glahaye, line 163: For completeness, add around line 80 that this endpoint can return 504

dehoward (approved):

glahaye (approved):

PR #305, opened 2023-09-06
Update SK from 0.19.230804.2-preview to 0.21.230828.2-preview

TaoChenOSU:
### Motivation and Context

Update SK from 0.19.230804.2-preview to 0.21.230828.2-preview

### Description

The following changes have been made because of this update:
1. Move away from using the logger (deprecated/removed) in SKcontext. Instead, it's going to use the logger that's passed in to the ChatSkill.
2. Planner will always retry regardless of the error type (when is configured) as a result of the removal of PlanningException and KernelException along with the error codes.

### Contribution Checklist

### DIFF
--- a/webapi/Options/PlannerOptions.cs
+++ b/webapi/Options/PlannerOptions.cs
@@ -13,7 +13,7 @@ public class PlannerOptions
     /// <summary>
     /// Options to handle planner errors.
     /// </summary>
-    public class PLannerErrorOptions
+    public class PlannerErrorOptions
     {
         /// <summary>
         /// Whether to allow retries on planner errors.
@@ -49,7 +49,7 @@ public class PlannerOptions
     /// <summary>
     /// Options on how to handle planner errors.
     /// </summary>
-    public PLannerErrorOptions PlannerError { get; set; } = new PLannerErrorOptions();
+    public PlannerErrorOptions PlannerError { get; set; } = new PlannerErrorOptions();
 
     /// <summary>
     /// The configuration for the stepwise planner.
--- a/webapi/Options/PlannerOptions.cs
+++ b/webapi/Options/PlannerOptions.cs
@@ -13,16 +13,18 @@ public class PlannerOptions
     /// <summary>
     /// Options to handle planner errors.
     /// </summary>
-    public class PlannerErrorOptions
+    public class ErrorOptions
     {
         /// <summary>
         /// Whether to allow retries on planner errors.
         /// </summary>
         public bool AllowRetries { get; set; } = true;
 
-        /// <summary>
-        /// Whether to allow missing functions.
-        /// </summary>
+        // <summary>
+        // Whether to allow missing functions in the sequential plan on creation. If set to true, the
+        // plan will be created with missing functions as no-op steps. If set to false (default),
+        // the plan creation will fail if any functions are missing.
+        // </summary>
         public bool AllowMissingFunctions { get; set; } = true;
 
         /// <summary>
@@ -49,7 +51,7 @@ public class PlannerOptions
     /// <summary>
     /// Options on how to handle planner errors.
     /// </summary>
-    public PlannerErrorOptions PlannerError { get; set; } = new PlannerErrorOptions();
+    public ErrorOptions ErrorHandling { get; set; } = new ErrorOptions();
 
     /// <summary>
     /// The configuration for the stepwise planner.
--- a/webapi/Options/PlannerOptions.cs
+++ b/webapi/Options/PlannerOptions.cs
@@ -11,20 +11,22 @@ namespace CopilotChat.WebApi.Options;
 public class PlannerOptions
 {
     /// <summary>
-    /// Whether to allow missing functions in plan on creation. If allowed, proposed plan will be sanitized of no-op functions.
-    /// Functions are considered missing if they're not available in the planner's kernel's context.
+    /// Options to handle planner errors.
     /// </summary>
-    public class MissingFunctionErrorOptions
+    public class PLannerErrorOptions
     {
         /// <summary>
-        /// Flag to indicate if skips are allowed on MissingFunction error
-        /// If set to true, the plan will be created with missing functions as no-op steps that are filtered from the final proposed plan.
-        /// If this is set to false, the plan creation will fail if any functions are missing.
+        /// Whether to allow retries on planner errors.
         /// </summary>
         public bool AllowRetries { get; set; } = true;
 
         /// <summary>
-        /// Max retries allowed on MissingFunctionsError.
+        /// Whether to allow missing functions.
+        /// </summary>
+        public bool AllowMissingFunctions { get; set; } = true;
+
+        /// <summary>
+        /// Max retries allowed.
         /// </summary>
         [Range(1, 5)]
         public int MaxRetriesAllowed { get; set; } = 3;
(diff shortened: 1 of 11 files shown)


alliscode (approved):

webapi/Options/PlannerOptions.cs
```diff
@@ -11,20 +11,22 @@ namespace CopilotChat.WebApi.Options;
 public class PlannerOptions
 {
     /// <summary>
-    /// Whether to allow missing functions in plan on creation. If allowed, proposed plan will be sanitized of no-op functions.
-    /// Functions are considered missing if they're not available in the planner's kernel's context.
+    /// Options to handle planner errors.
     /// </summary>
-    public class MissingFunctionErrorOptions
+    public class PLannerErrorOptions
```
dehoward, line 16:
```suggestion
    public class PlannerErrorOptions
```

dehoward (changes requested):

crickman (approved):

webapi/Options/PlannerOptions.cs
```diff
@@ -11,20 +11,22 @@ namespace CopilotChat.WebApi.Options;
...
-        /// Flag to indicate if skips are allowed on MissingFunction error
-        /// If set to true, the plan will be created with missing functions as no-op steps that are filtered from the final proposed plan.
-        /// If this is set to false, the plan creation will fail if any functions are missing.
+        /// Whether to allow retries on planner errors.
         /// </summary>
         public bool AllowRetries { get; set; } = true;
 
         /// <summary>
-        /// Max retries allowed on MissingFunctionsError.
+        /// Whether to allow missing functions.
```
glahaye, line 24: You probably want to keep the description of what this boolean does.  It's not obvious for people who don't know about it.

webapi/Skills/ChatSkills/CopilotChatPlanner.cs
```diff
@@ -95,7 +102,7 @@ public async Task<Plan> CreatePlanAsync(string goal, ILogger logger)
                         {
                             RelevancyThreshold = this._plannerOptions?.RelevancyThreshold,
                             // Allow plan to be created with missing functions
-                            AllowMissingFunctions = this._plannerOptions?.MissingFunctionError.AllowRetries ?? false
+                            AllowMissingFunctions = this._plannerOptions?.PlannerError.AllowMissingFunctions ?? false
```
glahaye, line 105: Nice!

webapi/Skills/ChatSkills/ExternalInformationSkill.cs
```diff
@@ -153,25 +159,20 @@ public class ExternalInformationSkill
             Plan? plan = null;
             // Use default planner options if planner options are null.
             var plannerOptions = this._planner.PlannerOptions ?? new PlannerOptions();
-            int retriesAvail = plannerOptions.MissingFunctionError.AllowRetries
-                ? plannerOptions.MissingFunctionError.MaxRetriesAllowed // Will always be at least 1
-                : plannerOptions.AllowRetriesOnInvalidPlan ? 1 : 0;
+            int retriesAvail = plannerOptions.PlannerError.AllowRetries
```
glahaye, line 162: Is this correct??

When we allow retries, retriesAvail is set to 0, otherwise it's MaxRetriesAllowed???

It feels like it should be the other way around.

If this is indeed correct, please put a comment to explain this logic

webapi/Skills/ChatSkills/ExternalInformationSkill.cs
```diff
@@ -204,25 +205,6 @@ public class ExternalInformationSkill
 
     #region Private
 
-    /// <summary>
```
glahaye, line 207: Should we really return an empty string at line 203??

webapi/appsettings.json
```diff
@@ -61,15 +61,11 @@
...
     "RelevancyThreshold": "0.25",
-    // Whether to allow missing functions in the plan on creation then sanitize output. Functions are considered missing if they're not available in the planner's kernel's context.
-    // If set to true, the plan will be created with missing functions as no-op steps that are filtered from the final proposed plan.
-    // If this is set to false, the plan creation will fail if any functions are missing.
-    "MissingFunctionsError": {
-      "AllowSkips": "true",
-      "MaxRetriesAllowed": "3" // Max retries allowed on MissingFunctionsError. If set to 0, no retries will be attempted.
+    "PlannerError": {
+      "AllowRetries": "true",
+      "AllowMissingFunctions": "true",
```
glahaye, line 66: Again, I believe we should keep a comment to explain what AllowMissingFunctions is

teresaqhoang: +1 on adding the comment. Would recommend renaming this to 
```suggestion
      "SkipMissingFunctionsError": "true",
```

webapi/Options/PlannerOptions.cs
```diff
@@ -11,20 +11,22 @@ namespace CopilotChat.WebApi.Options;
 public class PlannerOptions
 {
     /// <summary>
-    /// Whether to allow missing functions in plan on creation. If allowed, proposed plan will be sanitized of no-op functions.
-    /// Functions are considered missing if they're not available in the planner's kernel's context.
+    /// Options to handle planner errors.
     /// </summary>
-    public class MissingFunctionErrorOptions
+    public class PlannerErrorOptions
```
teresaqhoang, line 16: Planner is redundant here since the containing class is PlannerOptions

```suggestion
    public class ErrorOptions
```

webapi/Options/PlannerOptions.cs
```diff
@@ -45,14 +47,9 @@ public class MissingFunctionErrorOptions
...
-    /// Options on how to handle missing functions in plan on creation.
-    /// </summary>
-    public MissingFunctionErrorOptions MissingFunctionError { get; set; } = new MissingFunctionErrorOptions();
-
-    /// <summary>
-    /// Whether to retry plan creation if LLM returned response that doesn't contain valid plan (e.g., invalid XML or JSON, contains missing function, etc.).
+    /// Options on how to handle planner errors.
     /// </summary>
-    public bool AllowRetriesOnInvalidPlan { get; set; } = true;
+    public PlannerErrorOptions PlannerError { get; set; } = new PlannerErrorOptions();
```
teresaqhoang, line 52: Same note here
```suggestion
    public ErrorOptions ErrorHandling { get; set; } = new ErrorOptions();
```

webapi/Skills/ChatSkills/ExternalInformationSkill.cs
TaoChenOSU: Good catch!

webapi/Skills/ChatSkills/ExternalInformationSkill.cs
TaoChenOSU: Yes, this is intended.

webapi/appsettings.json
TaoChenOSU: `AllowMissingFunctions` is defined in the `SequentialPlannerConfig`. I think it's better to align the names. I will add a comment.

crickman (approved):

alliscode (approved):

webapi/appsettings.json
teresaqhoang: It's not the same behavior. In the `SequentialPlannerConfig`, AllowMissingFunctions allows the missing functions in the final plan it returns, whereas enabling this flag here means allowing the missing functions and then removing them from the final plan, so I think `Skip` is more indicative of what the user is setting here.

dehoward (approved):

PR #308, opened 2023-09-06
Migration POC

crickman:
### Motivation and Context

Break-out memory migration work for ease of review

### Description

One-time migration of all chat memory (working and long-term).

### Contribution Checklist

### DIFF
--- a/webapp/src/App.tsx
+++ b/webapp/src/App.tsx
@@ -63,11 +63,13 @@ const App: FC = () => {
     const dispatch = useAppDispatch();
 
     const { instance, inProgress } = useMsal();
-    const { activeUserInfo, features } = useAppSelector((state: RootState) => state.app);
+    const { activeUserInfo, features, isMigrating } = useAppSelector((state: RootState) => state.app);
     const isAuthenticated = useIsAuthenticated();
 
     const chat = useChat();
     const file = useFile();
+
+    console.log(`# ${isMigrating} (APP)`); // $$$
 
     useEffect(() => {
         if (isAuthenticated) {
@@ -125,6 +127,11 @@ const App: FC = () => {
             ]);
         }
 
+        if (isMigrating)
+        {
+            setAppState(AppState.ProbeForBackend);
+        }
+    
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [instance, inProgress, isAuthenticated, appState]);
 
--- a/webapp/src/components/views/BackendProbe.tsx
+++ b/webapp/src/components/views/BackendProbe.tsx
@@ -3,6 +3,8 @@
 import { Body1, Spinner, Title3 } from '@fluentui/react-components';
 import { FC, useEffect } from 'react';
 import { useSharedClasses } from '../../styles';
+import { useAppSelector } from '../../redux/app/hooks';
+import { RootState } from '../../redux/app/store';
 
 interface IData {
     uri: string;
@@ -11,20 +13,53 @@ interface IData {
 
 export const BackendProbe: FC<IData> = ({ uri, onBackendFound }) => {
     const classes = useSharedClasses();
+    const { isMigrating } = useAppSelector((state: RootState) => state.app);
+    const healthUrl = new URL('healthz', uri);
+    const migrationUrl = new URL('maintenance', uri);
+
+    console.log(`# ${isMigrating} (PROBE)`); // $$$
+
     useEffect(() => {
         const timer = setInterval(() => {
-            const requestUrl = new URL('healthz', uri);
-            const fetchAsync = async () => {
-                const result = await fetch(requestUrl);
+            const fetchHealthAsync = async () => {
+                const result = await fetch(healthUrl);
+
+                if (isMigrating) {
+                    console.log(`# MIGRATING`); // $$$
+                    return;
+                }
 
                 if (result.ok) {
+                    console.log(`# HEALTH DONE`); // $$$
+                    onBackendFound();
+                }
+
+                console.log(`# HEALTH NEXT`); // $$$
+            };
+
+            const fetchMigrationAsync = async () => {
+                const result = await fetch(migrationUrl);
+
+                if (!result.ok) {
+                    return;
+                }
+
+                const text = "test" + 1; // $$$
+
+                if (text === 'None') {
                     onBackendFound();
                 }
             };
 
-            fetchAsync().catch(() => {
+            fetchHealthAsync().catch(() => {
                 // Ignore - this page is just a probe, so we don't need to show any errors if backend is not found
             });
+
+            if (isMigrating) {
+                fetchMigrationAsync().catch(() => {
+                    // Ignore - this page is just a probe, so we don't need to show any errors if backend is not found
+                });
+            }
         }, 3000);
 
         return () => {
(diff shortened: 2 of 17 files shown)


webapp/src/App.tsx
```diff
@@ -125,6 +127,11 @@ const App: FC = () => {
             ]);
         }
 
+        if (isMigrating)
+        {
+            setAppState(AppState.ProbeForBackend);
+        }
+    
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [instance, inProgress, isAuthenticated, appState]);
```
teresaqhoang, line 136: isMigrating needs to be part of the dependency array.
If it's causing an infinite loop, try changing the check on line 130 to 
`if (isMigrating && appState !== AppState.ProbeForBackend)`

```suggestion
    }, [instance, inProgress, isAuthenticated, appState, isMigrating]);
```

webapp/src/components/views/BackendProbe.tsx
```diff
@@ -11,20 +13,53 @@ interface IData {
...
+
+                const text = "test" + 1; // $$$
+
+                if (text === 'None') {
                     onBackendFound();
                 }
             };
 
-            fetchAsync().catch(() => {
+            fetchHealthAsync().catch(() => {
```
teresaqhoang, line 54: Optimization since health will always be verified before migration check (optionally, you can put this in an else statement below)

```suggestion
            if (!isMigrating) {
                   fetchHealthAsync().catch(() => {
```

then you can also remove the isMigrating check in lines 27-30

webapi/Controllers/MaintenanceController.cs
```diff
@@ -46,14 +47,34 @@ public class MaintenanceController : ControllerBase
...
-    public ActionResult<MaintenanceResult?> GetMaintenanceStatus(
-        [FromServices] IKernel kernel,
+    public async Task<ActionResult<MaintenanceResult?>> GetMaintenanceStatusAsync(
+        [FromServices] IChatMigrationMonitor migrationMonitor,
         [FromServices] IHubContext<MessageRelayHub> messageRelayHubContext,
         CancellationToken cancellationToken = default)
     {
+        MaintenanceResult? result = null;
+
+        var migrationStatus = await migrationMonitor.GetCurrentStatusAsync(cancellationToken).ConfigureAwait(false);
```
alliscode, line 57: nit: ASP .net core doesnt have a SynchronizationContext so ConfigureAwait is not needed.

webapi/Services/MemoryMigration/ChatMemoryMigrationService.cs
```diff
@@ -0,0 +1,146 @@
...
+        {
+            var chats = await this._chatSessionRepository.GetAllChatsAsync().ConfigureAwait(false);
+            foreach (var chat in chats)
+            {
+                foreach (var memoryType in this._promptOptions.MemoryMap.Keys)
+                {
+                    var indexName = $"{chat.Id}-{memoryType}";
+                    if (collectionNames.Contains(indexName))
+                    {
+                        var memories = await this._memory.SearchAsync(indexName, "*", limit: 10000, minRelevanceScore: 0, withEmbeddings: false, cancellationToken).ToArrayAsync(cancellationToken);
```
TaoChenOSU, line 109: Is 1000 enough?

webapi/Services/MemoryMigration/ChatMemoryMigrationService.cs
```diff
@@ -0,0 +1,146 @@
...
+            return;
+        }
+
+        //  Create memory token
+        var token = Guid.NewGuid().ToString();
+        await SetTokenMemory(token, cancellationToken).ConfigureAwait(false);
+
+        await RemoveMemorySourcesAsync().ConfigureAwait(false);
+
+        bool needsZombie = true;
```
alliscode, line 83: A comment describing the role of  `needsZombie` would be helpful.

webapi/Services/MemoryMigration/ChatMigrationMaintenanceAction.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        var migrationStatus = await this._migrationMonitor.GetCurrentStatusAsync(cancellation).ConfigureAwait(false);
+
+        switch (migrationStatus)
+        {
+            case ChatMigrationStatus s when (s == ChatMigrationStatus.RequiresUpgrade):
+                try
+                {
+                    // Migrate all chats to single index (in background)
+                    var task = this._migrationService.MigrateAsync(cancellation);
```
alliscode, line 40: It looks like `task` is never used? Also, I believe the MIgrateAsync call will not throw until the task is awaited so this catch block will not catch anything.

webapi/Services/MemoryMigration/ChatMigrationStatus.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace CopilotChat.WebApi.Services.MemoryMigration;
+
+/// <summary>
+/// Set of migration states/status for chat memory migration.
+/// </summary>
+/// <remarks>
+/// Interlocked.CompareExchange doesn't work with enums.
+/// </remarks>
+public sealed class ChatMigrationStatus
```
TaoChenOSU, line 11: nit: Would a enum work>

webapi/Controllers/MaintenanceController.cs
alliscode: Same comment for elsewhere in this PR.

crickman: Ah great, nice to know!

webapi/Services/MemoryMigration/ChatMemoryMigrationService.cs
crickman: No doubt...good point.

webapi/Services/MemoryMigration/ChatMemoryMigrationService.cs
crickman: We don't know what the customer may have, but 10000 is the maximum allowed value (anything higher than that results in a fail-fast exception).

webapi/Services/MemoryMigration/ChatMigrationStatus.cs
crickman: That was my starting point.  Apparently, Interlocked.CompareExchange doesn't work with enums.

webapi/Controllers/MaintenanceController.cs
crickman: Updated

webapi/Services/MemoryMigration/ChatMemoryMigrationService.cs
crickman: Updated

webapi/Services/MemoryMigration/ChatMigrationMaintenanceAction.cs
crickman: Task is required so satisfy analyzer.  Yeah try/catch is left-over - gone!

alliscode (approved):

TaoChenOSU (approved):

PR #316, opened 2023-09-08
Maintenance framework

crickman:
### Motivation and Context

In order to support maintenance modes that might disrupt service availability (such as a storage update), adding support to put app into a maintenance mode.

### Description
image
image

### Contribution Checklist

### DIFF
--- a/webapi/Models/Response/MigrationResult.cs
+++ b/webapi/Models/Response/MigrationResult.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace CopilotChat.WebApi.Models.Response;
+
+/// <summary>
+/// $$$
+/// </summary>
+public class MigrationResult
+{
+    /// <summary>
+    /// $$$
+    /// </summary>
+    public string Title { get; set; } = string.Empty;
+
+    /// <summary>
+    /// $$$
+    /// </summary>
+    public string Messgae { get; set; } = string.Empty;
+
+    /// <summary>
+    /// $$$
+    /// </summary>
+    public string? Note { get; set; }
+}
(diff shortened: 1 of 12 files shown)


webapi/Models/Response/MigrationResult.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// <summary>
+    /// $$$
+    /// </summary>
+    public string Message { get; set; } = string.Empty;
+
+    /// <summary>
+    /// $$$
+    /// </summary>
+    public string? Note { get; set; }
```
alliscode, line 23: Do these ever get set?

webapi/appsettings.json
```diff
@@ -15,11 +15,13 @@
   // - Optionally set TimeoutLimitInS to the maximum number of seconds to wait for a response from the AI service. If this is not set, there is no timeout.
   // - Optionally set SemanticSkillsDirectory to the directory from which to load semantic skills (e.g., "./SemanticSkills").
   // - Optionally set KeyVaultUri to the URI of the Key Vault for secrets (e.g., "https://contoso.vault.azure.net/").
+  // - Optionally set InMaintenance to true to set the application to maintenance mode.
   //
   "Service": {
     // "TimeoutLimitInS": "120"
     // "SemanticSkillsDirectory": "",
     // "KeyVault": ""
+    // "InMaintenance":  true
```
alliscode, line 24: If this is set to true, how does the app come out of maintenance mode without redeploying it?

webapi/Models/Response/MigrationResult.cs
crickman: Not in this PR.  They will for the memory migration...otherwise the defaults are used.

webapi/appsettings.json
crickman: Set to false.  Using the config like this would be for something like the partition upgrade (external to the api service).

With the memory-migration, it will dynamically indicate maintenance as an overlay to this config (or condition)

alliscode (approved):

webapi/appsettings.json
alliscode: Discussed offline, I get it now, thanks!

PR #336, opened 2023-09-12
Fix multi-user chat feature toggle behavior

TaoChenOSU:
### Motivation and Context

A couple of issues:
1. Multi-user chat sessions are filtered out in the loadChats method if the feature is disabled. When the feature is enabled, it doesn't call loadChats again, making it impossible to retrieve multi-user chat sessions. 
2. Supposed a user has only multi-user chat session, when the app first starts, all the multi-user chat sessions will be filtered out but the app doesn't call create chat, leaving the user with no chat to interact with.
3. the initial selectedId is the id of the first chat session returned by the service. If the first chat session is a multi-user chat session, the app will crash because the first chat session is filtered out and the selectedId becomes invalid.

### Description
Address all the above.

### Contribution Checklist

### DIFF
--- a/webapp/src/components/chat/chat-list/ChatList.tsx
+++ b/webapp/src/components/chat/chat-list/ChatList.tsx
@@ -133,18 +133,25 @@ export const ChatList: FC = () => {
 
     useEffect(() => {
         // Ensure local component state is in line with app state.
+        const nonHiddenConversations: Conversations = {};
+        for (const key in conversations) {
+            if (!conversations[key].hidden) {
+                nonHiddenConversations[key] = conversations[key];
+            }
+        }
+
         if (filterText !== '') {
             // Reapply search string to the updated conversations list.
             const filteredConversations: Conversations = {};
-            for (const key in conversations) {
-                if (conversations[key].title.toLowerCase().includes(filterText.toLowerCase())) {
-                    filteredConversations[key] = conversations[key];
+            for (const key in nonHiddenConversations) {
+                if (nonHiddenConversations[key].title.toLowerCase().includes(filterText.toLowerCase())) {
+                    filteredConversations[key] = nonHiddenConversations[key];
                 }
             }
             setConversationsView({ filteredConversations: filteredConversations });
         } else {
             // If no search string, show full conversations list.
-            setConversationsView(sortConversations(conversations));
+            setConversationsView(sortConversations(nonHiddenConversations));
         }
     }, [conversations, filterText]);
 
--- a/webapp/src/components/chat/chat-list/ChatList.tsx
+++ b/webapp/src/components/chat/chat-list/ChatList.tsx
@@ -133,26 +133,15 @@ export const ChatList: FC = () => {
 
     useEffect(() => {
         // Ensure local component state is in line with app state.
-        const nonHiddenConversations: Conversations = {};
+        const filteredConversations: Conversations = {};
         for (const key in conversations) {
-            if (!conversations[key].hidden) {
-                nonHiddenConversations[key] = conversations[key];
+            const conversation = conversations[key];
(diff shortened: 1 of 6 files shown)


webapp/src/components/chat/chat-list/ChatList.tsx
```diff
@@ -133,18 +133,25 @@ export const ChatList: FC = () => {
...
+                if (nonHiddenConversations[key].title.toLowerCase().includes(filterText.toLowerCase())) {
+                    filteredConversations[key] = nonHiddenConversations[key];
                 }
             }
             setConversationsView({ filteredConversations: filteredConversations });
         } else {
             // If no search string, show full conversations list.
-            setConversationsView(sortConversations(conversations));
+            setConversationsView(sortConversations(nonHiddenConversations));
         }
```
dehoward, line 155: we can cut down on the second loop by combining the conditions:

```suggestion
        const filteredConversations: Conversations = {};
        for (const key in conversations) {
            const conversation = conversations[key];
            if (
                !conversation.hidden &&
                (!filterText || conversation.title.toLowerCase().includes(filterText.toLowerCase()))
            ) {
                filteredConversations[key] = conversation;
            }
        }

        setConversationsView(filterText ? { filteredConversations } : sortConversations(filteredConversations));
```

webapp/src/components/header/settings-dialog/SettingSection.tsx
```diff
@@ -30,6 +31,9 @@ export const SettingSection: React.FC<ISettingsSectionProps> = ({ setting, conte
     const onFeatureChange = useCallback(
         (featureKey: FeatureKeys) => {
             dispatch(toggleFeatureFlag(featureKey));
+            if (featureKey === FeatureKeys.MultiUserChat) {
+                dispatch(toggleMultiUserConversations({}));
```
dehoward, line 35: remove as this is not an arg

do you mind also removing the `Slice<ConversationsState>` typing on line 16 in `conversationsSlice`? this is what is allowing passing in incorrect args

dehoward (changes requested):

webapp/src/components/chat/chat-list/ChatList.tsx
```diff
@@ -133,19 +133,15 @@ export const ChatList: FC = () => {
...
     useEffect(() => {
         // Ensure local component state is in line with app state.
-        if (filterText !== '') {
-            // Reapply search string to the updated conversations list.
-            const filteredConversations: Conversations = {};
-            for (const key in conversations) {
-                if (conversations[key].title.toLowerCase().includes(filterText.toLowerCase())) {
-                    filteredConversations[key] = conversations[key];
-                }
+        const filteredConversations: Conversations = {};
```
glahaye, line 136: These are conversations that are filtered IN, right?  Not conversations that are filtered OUT?

If so, a better name might be conversationsToDisplay or something similar since filteredConversations seems to imply that the conversations are filtered OUT, which is misleading.

webapp/src/libs/hooks/useChat.ts
```diff
@@ -169,11 +165,19 @@ export const useChat = () => {
                         botResponseStatus: undefined,
                         userDataLoaded: false,
                         disabled: false,
+                        hidden: !features[FeatureKeys.MultiUserChat].enabled && chatUsers.length > 1,
```
glahaye, line 168: I'm not sure I get the logic...  We hide this if multiuser is enabled or if chatUsers.length is equal to 1?

TaoChenOSU: We hide chats that have multiple users only when the feature is not enabled.

glahaye (approved):

dehoward (approved):

PR #337, opened 2023-09-12
Refactor ChatSkill to use new Chat Completions prompt format

teresaqhoang:
### Motivation and Context

This PR refactors the ChatSkill to use the[ new ChatCompletions prompt format](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/chatgpt?pivots=programming-language-chat-completions#system-role) as the meta-prompt template. The new format serves the prompt template as a series of context messages, where each message can be tagged a role, to prime the model in a certain way. These roles include: System, User, and Assistant.

### Description

For ChatSkill, the prompt template for the meta-bot response request utilizes the message roles in the following way:
- System: for any instructions or additional context provided to the model, including bot persona, relevant memories, or planner output. 
- Assistant: Any bot response message in chat history previously outputted as chat completion result.
- User: Any user input.

The series of messages between the user and the assistant is included both as few shot examples and additional context (chat history). The last user message is always included.

Other changes:
- This PR also adds a custom token calculator, since chat streaming doesn't return token usage details. 
- System continuation preamble is no longer needed with this new format, since it's been optimized for chat responses.
- Added visible scroll bar to prompt dialog.
- Simplified SystemResponse since this conversational interface is already optimized for chat.

image
image

### Contribution Checklist

~- [ ] All unit tests pass, and I have added new tests where possible~


webapi/Models/Response/BotResponsePrompt.cs
```diff
@@ -46,19 +47,13 @@ public class BotResponsePrompt
...
-    [JsonPropertyName("systemChatContinuation")]
-    public string SystemChatContinuation { get; set; } = string.Empty;
-
-    /// <summary>
-    /// Raw content of the rendered prompt.
-    /// </summary>
-    [JsonPropertyName("rawContent")]
-    public string RawContent { get; set; } = string.Empty;
+    [JsonPropertyName("metaPromptTemplate")]
+    public ChatHistory MetaPromptTemplate { get; set; } = new();
```
TaoChenOSU, line 54: nit:
The name of this type makes the code slightly confusing.
Suggestion:
Replace `using Microsoft.SemanticKernel.AI.ChatCompletion;` above to `using CompletionModelChatHistory = Microsoft.SemanticKernel.AI.ChatCompletion.ChatHistory;` 
Replace `public ChatHistory MetaPromptTemplate { get; set; } = new();` to `public CompletionModelChatHistory MetaPromptTemplate { get; set; } = new();`

webapi/Models/Response/BotResponsePrompt.cs
```diff
@@ -67,16 +62,15 @@ public class BotResponsePrompt
         string documentMemories,
         SemanticDependency<StepwiseThoughtProcess> externalInformation,
         string chatHistory,
-        string systemChatContinuation
-        )
+        ChatHistory metaPromptTemplate
```
TaoChenOSU, line 65: Same. `ChatHistory` is a type and variable name at the same time. Makes it a bit hard to understand at the first glance.

webapi/Skills/TokenUtilities.cs
```diff
@@ -87,4 +88,27 @@ internal static void GetFunctionTokenUsage(SKContext result, SKContext chatConte
     /// Calculate the number of tokens in a string.
     /// </summary>
     internal static int TokenCount(string text) => GPT3Tokenizer.Encode(text).Count;
+
+    /// <summary>
+    /// Rough token costing of ChatHistory's message object.
+    /// </summary>
+    internal static int GetChatMessageTokenCount(AuthorRole authorRole, string content)
+    {
+        var tokenCount = authorRole == AuthorRole.System ? TokenCount("\n") : 0;
```
TaoChenOSU, line 97: bit: comments/documentation on the logic behind this statement

webapp/src/components/utils/TextUtils.tsx
```diff
@@ -51,7 +51,7 @@ export function formatChatTextContent(messageContent: string) {
 /*
  * Formats text containing `\n` or `\r` into paragraphs.
  */
-export function formatParagraphTextContent(messageContent: string) {
+export function formatParagraphTextContent(messageContent = '') {
```
glahaye, line 54: Wouldn't the following be more explicit about the type:

```suggestion
export function formatParagraphTextContent(messageContent: string = '') {
```

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -215,47 +216,82 @@ public async Task<string> ExtractAudienceAsync(SKContext context, CancellationTo
...
     /// <param name="cancellationToken">The cancellation token.</param>
     [SKFunction, Description("Extract chat history")]
-    public async Task<string> ExtractChatHistoryAsync(
+    public async Task<string> ExtractChatHistoryAsStringAsync(
         [Description("Chat ID to extract history from")] string chatId,
         [Description("Maximum number of tokens")] int tokenLimit,
         CancellationToken cancellationToken = default)
+    {
+        return await this.ExtractChatHistoryAsync(chatId, tokenLimit, cancellationToken: cancellationToken);
+    }
```
glahaye, line 227: Feel free to use this or not but I just wanted to let you know you can also just do the following:

```suggestion
    public Task<string> ExtractChatHistoryAsStringAsync(string chatId, int tokenLimit,
        CancellationToken cancellationToken = default)
    {
        return this.ExtractChatHistoryAsync(chatId, tokenLimit, cancellationToken: cancellationToken);
    }
```

This just makes the caller await instead of us awaiting and then the caller await on us. It's pretty much the same except for where exceptions would be thrown.

webapp/src/components/utils/TextUtils.tsx
teresaqhoang: type isn't needed here since it can be inferred

`Type string trivially inferred from a string literal, remove type annotation.eslint[@typescript-eslint/no-inferrable-types](https://typescript-eslint.io/rules/no-inferrable-types)`

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: I see! This method is called from a semantic function (so in a rendered prompt). I'm not sure what behavior is there, if the caller awaits async functions or not. Do you know? And would it change your suggestion?

glahaye: It does not make a difference for the callers.  Either way, the caller will have to await.

The only difference is that in one case, they'll await on the inner call's Task passed straight through ExtractChatHistoryAsStringAsync() and in the other case, ExtractChatHistoryAsStringAsync() is the one that awaits the inner call while the called awaits ExtractChatHistoryAsStringAsync().

It's not much of a difference either way.  I just wanted you to be aware that either can be done.

It's mostly a matter for exception handling.

For more info:
https://code-maze.com/charp-difference-between-returning-and-awaiting-a-task/

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -215,47 +216,82 @@ public async Task<string> ExtractAudienceAsync(SKContext context, CancellationTo
...
+
+    /// <summary>
+    /// Extract chat history.
+    /// </summary>
+    /// <param name="chatId">Chat ID to extract history from.</param>
+    /// <param name="tokenLimit">Maximum number of tokens.</param>
+    /// <param name="chatHistory">ChatHistory object tracking allotted messages.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    private async Task<string> ExtractChatHistoryAsync(
+        [Description("Chat ID to extract history from")] string chatId,
```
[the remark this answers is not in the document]
teresaqhoang: Good catch, removed

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -215,47 +216,82 @@ public async Task<string> ExtractAudienceAsync(SKContext context, CancellationTo
     /// <summary>
     /// Extract chat history.
     /// </summary>
-    /// <param name="context">Contains the 'tokenLimit' controlling the length of the prompt.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     [SKFunction, Description("Extract chat history")]
-    public async Task<string> ExtractChatHistoryAsync(
+    public async Task<string> ExtractChatHistoryAsStringAsync(
```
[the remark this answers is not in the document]
teresaqhoang: Renamed this to `ExtractChatHistory` and the other one to `GetAllowedChatHistoryAsync`. 

I believe SK functions don't allow non-native types as parameters, so ChatHistory isn't a valid parameter type, so this wrapper is required so that `ExtractChatHistory` can be called from a rendered prompt.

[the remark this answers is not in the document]
teresaqhoang: Also added clarifying comments to `ExtractChatHistory` description.

webapi/Models/Response/BotResponsePrompt.cs
teresaqhoang: Totally agree, this was also bugging me in development. Renamed all cases of `Microsoft.SemanticKernel.AI.ChatCompletion.ChatHistory` to `ChatCompletionContextMessages` for clarity

webapi/Models/Response/BotResponsePrompt.cs
teresaqhoang: Renamed messages used for meta-prompt template to use `ChatCompletionContextMessages` type and `contextMessages` as variable name where applicable

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: Thanks for explaining! I changed it per your request since we mostly let exceptions bubble up anyway.

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -394,59 +446,42 @@ private async Task<ChatMessage> GetChatResponseAsync(string chatId, string userI
...
-        {
-            await this.UpdateBotResponseStatusOnClientAsync(chatId, "Extracting chat history", cancellationToken);
-            chatHistory = await this.ExtractChatHistoryAsync(chatId, chatHistoryTokenLimit, cancellationToken);
-            chatContext.ThrowIfFailed();
-            chatContextText = $"{chatContextText}\n{chatHistory}";
-        }
+        // Append previous messages
+        await this.UpdateBotResponseStatusOnClientAsync(chatId, "Extracting chat history", cancellationToken);
+        chatHistory = await this.GetAllowedChatHistoryAsync(chatId, chatHistoryTokenBudget, promptTemplate, cancellationToken);
+        chatContext.ThrowIfFailed();
```
TaoChenOSU, line 475: Do we need this?

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -394,59 +446,42 @@ private async Task<ChatMessage> GetChatResponseAsync(string chatId, string userI
...
-        // Need to extract this from the rendered prompt because Time and Date are calculated during render
-        var systemChatContinuation = Regex.Match(renderedPrompt, PromptsOptions.SYSTEM_CHAT_CONTINUATION_REGEX).Value;
-        var promptView = new BotResponsePrompt(renderedPrompt, this._promptOptions.SystemDescription, this._promptOptions.SystemResponse, audience, userIntent, chatMemories, documentMemories, plannerDetails, chatHistory, systemChatContinuation);
-
-        // Calculate token usage of prompt template
-        chatContext.Variables.Set(TokenUtilities.GetFunctionKey(this._logger, "SystemMetaPrompt")!, TokenUtilities.TokenCount(renderedPrompt).ToString(CultureInfo.InvariantCulture));
-
+        var promptView = new BotResponsePrompt(this._promptOptions.SystemDescription, this._promptOptions.SystemResponse, audience, userIntent, chatMemories, documentMemories, plannerDetails, chatHistory, promptTemplate);
+        chatContext.Variables.Set(TokenUtilities.GetFunctionKey(this._logger, "SystemMetaPrompt")!, TokenUtilities.GetContextMessagesTokenCount(promptTemplate).ToString(CultureInfo.InvariantCulture));
         chatContext.ThrowIfFailed();
```
TaoChenOSU, line 485: Do we need this?

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -394,59 +446,42 @@ private async Task<ChatMessage> GetChatResponseAsync(string chatId, string userI
...
-        chatContext.Variables.Set("prompt", renderedPrompt);
-
-        // Need to extract this from the rendered prompt because Time and Date are calculated during render
-        var systemChatContinuation = Regex.Match(renderedPrompt, PromptsOptions.SYSTEM_CHAT_CONTINUATION_REGEX).Value;
-        var promptView = new BotResponsePrompt(renderedPrompt, this._promptOptions.SystemDescription, this._promptOptions.SystemResponse, audience, userIntent, chatMemories, documentMemories, plannerDetails, chatHistory, systemChatContinuation);
-
-        // Calculate token usage of prompt template
-        chatContext.Variables.Set(TokenUtilities.GetFunctionKey(this._logger, "SystemMetaPrompt")!, TokenUtilities.TokenCount(renderedPrompt).ToString(CultureInfo.InvariantCulture));
-
+        var promptView = new BotResponsePrompt(this._promptOptions.SystemDescription, this._promptOptions.SystemResponse, audience, userIntent, chatMemories, documentMemories, plannerDetails, chatHistory, promptTemplate);
```
TaoChenOSU, line 483: Looks like the prompt template is the object that's passed to the model and it contains everything the BotResponsePrompt contains. Why do we still need the BosResponsePrompt object?

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -213,49 +215,88 @@ public async Task<string> ExtractAudienceAsync(SKContext context, CancellationTo
...
+    /// </summary>
+    /// <param name="chatId">Chat ID to extract history from.</param>
+    /// <param name="tokenLimit">Maximum number of tokens.</param>
+    /// <param name="chatHistory">Optional ChatCompletionContextMessages object tracking allotted messages.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>Chat history as a string.</returns>
+    private async Task<string> GetAllowedChatHistoryAsync(
+        string chatId,
+        int tokenLimit,
+        ChatCompletionContextMessages? chatHistory = null,
```
TaoChenOSU, line 243: Why is this optional? This function can have two outputs depending on if this is passed in.

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: Need what?

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: Need what? Setting the variable in the chatContext?

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: To show the formatted sections in the UX. Otherwise, there's not a clean way to determine each component of the prompt. The prompt template is shown as the raw content
image

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: I'm not sure I know what you're talking about. 

This is optional because we don't need to update a ChatHistory object when extracting chat history as a string during other steps like audience and user intent extraction, and it's the same logic to allocate the messages. But we can't call this method directly because SK functions don't allow non-native types as parameters, so ChatHistory isn't a valid parameter type. The wrapper `ExtractChatHistory` above is required so that `GetAllowedChatHistoryAsync` can be called from a rendered prompt.

webapi/Skills/ChatSkills/ChatSkill.cs
TaoChenOSU: `chatContext.ThrowIfFailed();` None of the operations above touch the context.

webapi/Skills/ChatSkills/ChatSkill.cs
TaoChenOSU: Same as above.

webapi/Skills/ChatSkills/ChatSkill.cs
TaoChenOSU: Got it!

webapi/Skills/ChatSkills/ChatSkill.cs
TaoChenOSU: Aren't we using the chat completion model in user intent extraction and other places? Why is ChatHistory not needed?

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -339,30 +383,41 @@ public async Task<string> ExtractAudienceAsync(SKContext context, CancellationTo
...
     /// <returns>The created chat message containing the model-generated response.</returns>
-    private async Task<ChatMessage> GetChatResponseAsync(string chatId, string userId, SKContext chatContext, CancellationToken cancellationToken)
+    private async Task<ChatMessage> GetChatResponseAsync(string chatId, string userId, SKContext chatContext, ChatMessage userMessage, CancellationToken cancellationToken)
     {
+        // Render system instruction components
+        var promptRenderer = new PromptTemplateEngine();
+        var systemInstructions = await promptRenderer.RenderAsync(
+            this._promptOptions.SystemPersona,
+            chatContext,
+            cancellationToken);
```
dehoward, line 396: can we add a status update for this async call?
```suggestion
        await this.UpdateBotResponseStatusOnClient(chatId, "Rendering prompt template", cancellationToken);
        var promptRenderer = new PromptTemplateEngine();
        var systemInstructions = await promptRenderer.RenderAsync(
            this._promptOptions.SystemPersona,
            chatContext,
            cancellationToken);
```

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: Oh you're right, removed

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: removed

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: @TaoChenOSU We are using the chat completion model for all the initial hops to the LLM, but I opted out of refactoring those into the new prompt template format because they're not chat centric i.e., they don't need to be optimized for chat since the prompt template and user input for these don't demand a conversational bot response. 

Hence, it makes more sense to keep these in the format of text completion than chat completion, so ChatHistory isn't needed.

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: Added an update with status "Initializing prompt".

TaoChenOSU (approved):

dehoward (approved):

PR #346, opened 2023-09-13
Querying chat messages with partition key

teresaqhoang:
### Motivation and Context

Some changes to complement [partition key change](https://github.com/microsoft/chat-copilot/pull/240/files#diff-f9c37b8a51287ee8b7d42811b02d2ee33e4bcb565a155b358216519494eb38a9):
- Documentation updates to make partition key requirements more clear
- Using chatId as partition key when updating chat message state 
- Using partition key on upsert

### Contribution Checklist

~- [ ] All unit tests pass, and I have added new tests where possible~

### DIFF
--- a/webapi/Skills/ChatSkills/ChatSkill.cs
+++ b/webapi/Skills/ChatSkills/ChatSkill.cs
@@ -639,10 +639,13 @@ public class ChatSkill
     /// <param name="cancellationToken">The cancellation token.</param>
     private async Task UpdateChatMessageContentAsync(string updatedResponse, string messageId, string chatId, CancellationToken cancellationToken)
     {
-        // Make sure the chat exists.
-        var chatMessage = await this._chatMessageRepository.FindByIdAsync(messageId, chatId);
-        chatMessage.Content = updatedResponse;
+        ChatMessage? chatMessage = null;
+        if (!await this._chatMessageRepository.TryFindByIdAsync(messageId, chatId, callback: v => chatMessage = v))
+        {
+            throw new ArgumentException($"Chat message {messageId} does not exist.");
+        }
 
+        chatMessage!.Content = updatedResponse;
         await this._chatMessageRepository.UpsertAsync(chatMessage);
     }
 
--- a/webapi/Skills/ChatSkills/ChatSkill.cs
+++ b/webapi/Skills/ChatSkills/ChatSkill.cs
@@ -302,7 +302,7 @@ public class ChatSkill
         if (!string.IsNullOrWhiteSpace(planJson) &&
             !string.IsNullOrEmpty(messageId))
         {
-            await this.UpdateChatMessageContentAsync(planJson, messageId, cancellationToken);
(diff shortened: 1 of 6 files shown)


crickman (dismissed):

webapi/Storage/CosmosDbContext.cs
```diff
@@ -100,7 +100,7 @@ public async Task UpsertAsync(T entity)
             throw new ArgumentOutOfRangeException(nameof(entity.Id), "Entity Id cannot be null or empty.");
         }
 
-        await this._container.UpsertItemAsync(entity);
+        await this._container.UpsertItemAsync(entity, new PartitionKey(entity.Partition));
```
crickman, line 103: I thought the partition key was automatically extracted from the entity in this case.  It certainly doesn't hurt to be explicit.

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -634,14 +634,18 @@ private async Task<ChatMessage> SaveNewResponseAsync(string response, string pro
...
-        var chatMessage = await this._chatMessageRepository.FindByIdAsync(messageId);
-        chatMessage.Content = updatedResponse;
+        ChatMessage? chatMessage = null;
+        if (!await this._chatMessageRepository.TryFindByIdAsync(messageId, chatId, callback: v => chatMessage = v))
+        {
+            throw new ArgumentException($"Chat message {messageId} does not exist.");
+        }
 
+        chatMessage!.Content = updatedResponse;
         await this._chatMessageRepository.UpsertAsync(chatMessage);
```
crickman, line 649: Aux: This upsert isn't including explicit partition key

crickman (approved):

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: Since we're updating an existing entity, the partitionKey should already exist on the ChatMessage object and extracted in CosmosDbContext

webapi/Storage/CosmosDbContext.cs
teresaqhoang: Yeah it is, but this is mostly to align our sanity check with the new partitionKeys, i.e, before update would work with Cosmos containers configured before the new key requirements because it was auto-extracting the default partition key rather than throwing an error when trying to reference the new key value

glahaye (approved):

crickman (approved):

webapi/Skills/ChatSkills/ChatSkill.cs
crickman: FWIW - Upsert works the same way wrt partition key extraction for create or update (because the entire entity is present)

teresaqhoang: Totally, that's why I added the partition key to the call explicitly in CosmosDBContext. I'll add it to the Create call there for consistency too

glahaye (approved):
Ship it!

crickman (approved):

Issue #351, opened 2023-09-15
when use memory store qdrant , occurs 404 error


AwesomeYuer:
**Describe the bug**

This issue is easy for occurs! I remember the old version have not this 404 error issue

I am not sure this issue whether cause the issue https://github.com/microsoft/chat-copilot/issues/213#issue-1856745720

BTW:  Up to now, I found same issue as #213 is exists indeed, but it have been closed.

@crickman  have this #213 bug been fixed? or in fixing? thanks

**To Reproduce**
You need to config as below only, and then chat!

```json
{
  "MemoryStore": {
    "Type": "qdrant",
    "Qdrant": {
      "Host": "http://localhost",
      "Port": "6333",
      "VectorSize": 1536
      // "Key":  ""
    },
}
```

**Expected behavior**

**Screenshots**

![image](https://github.com/microsoft/chat-copilot/assets/1026479/f6d55b0d-05f7-4e50-baed-f7ae45f36b74)

**Platform**
 - OS: Windows
 - IDE: VS 2022
 - Language: [e.g. C#,]
 - Source: Latest code on 16th Sep, 2023

**Additional context**

crickman:
Hi AwesomeYuer, thank you for reporting this.   This is a known issue with all semantic-kernel connectors.  Chat-Copilot _should_ catche these and interprets the result appropriately.

Are you observing a functional failure?  If so, can you please provide additional information?

crickman:
> This seems to be present to all memory store, (I also have it with CognitiveSearch and someone else with PostgreSQL).
> 
> It should be fixed with #152

Indeed, we are moving away from SK connectors in this change!  Thank for for calling this out.

kbeaugrand:
Hi,

one question regarding this PR. By moving to semantic-memory, you remove the possibility to add new connectors directly to chat-copilot right?
We would only be able to use already implemented connectors in semantic-memory, if we want to leverage on other database, third-party connectors might be difficult to add in that case...

crickman:
Ok, I think I found the root issue, the exception is causing SaveInformationAsync to be skipped.

AwesomeYuer:
@crickman thanks.

How about #213? have this issue been fixed? or in fixing? thanks

Up to now, I uploaded a document via the UI, but the chatbot cannot answer questions on the document indeed.

Issue #215, opened 2023-08-18
Bug: The CopilotChatWebApi backend should invoke CreateCollectionAsync once at least for each chat session when use PostgresMemoryStore and local postgresql database as vectors MemoryStore


AwesomeYuer:
**Describe the bug**

The WebAPI backend should invoke CreateCollectionAsync once at least for each chat session when use PostgresMemoryStore and local postgresql database  as vectors MemoryStore

**To Reproduce**
1. running local postgres server
2. modify `CopilotChatWebApi` appsettings.json
```
"MemoryStore": {
  "Type": "postgres",
  "Qdrant": {
    "Host": "http://localhost",
    "Port": "6333",
    "VectorSize": 1536
    // "Key":  ""
  },
  "AzureCognitiveSearch": {
    "Endpoint": ""
    // "Key": ""
  },
  "Chroma": {
    "Host": "http://localhost",
    "Port": "8000"
  },
  "Postgres": {
    "VectorSize": 1536
    // "ConnectionString": // dotnet user-secrets set "MemoryStore:Postgres:ConnectionString" "MY_POSTGRES_CONNECTION_STRING"
  }
},
```
3. modify CopilotChatWebApi user secrets.json
```
{
  "AIService:Key": "eeXXXXXXXXXXXXXXXXXXXXXXX",
  "MemoryStore:Postgres:ConnectionString": "Host=localhost;Database=postgres;User Id=postgres;Password=password01!"
}
```
4. Debug CopilotChatWebApi in VS2022
5. Run webapp in VSCode using command
```
yarn serve
```
6. chat in new session in webapp UI
7. see same error in webapp frontend and webapi backend console
```
fail: Microsoft.SemanticKernel.IKernel[0]
      Something went wrong while executing the native function. Function: <GetDelegateInfo>b__0. Error: 42P01: relation "public.d8ff415b-167b-4b1a-8dd8-7d62f97f120d-LongTermMemory" does not exist

      POSITION: 115
      Npgsql.PostgresException (0x80004005): 42P01: relation "public.d8ff415b-167b-4b1a-8dd8-7d62f97f120d-LongTermMemory" does not exist

      POSITION: 115
         at Npgsql.Internal.NpgsqlConnector.<ReadMessage>g__ReadMessageLong|221_0(NpgsqlConnector connector, Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
[...]

PR #365, opened 2023-09-19
Maintenance and memory fixes

crickman:
### Motivation and Context

Testing [the ](https://github.com/microsoft/chat-copilot/pull/152), discovered:

1. Maintenance stabilization around state changes and controller result
2. Addressed root-cause for connecter issue reported in: https://github.com/microsoft/chat-copilot/issues/351 and https://github.com/microsoft/chat-copilot/issues/215

### Description

- Decouple exception handling for `SearchAsync` and `SaveInformationAsync`
- Change from "ref" to "state" handling for BackendProbe
- Refine handling of maintenance response

### Contribution Checklist

### DIFF
--- a/webapi/Controllers/MaintenanceController.cs
+++ b/webapi/Controllers/MaintenanceController.cs
@@ -55,9 +55,9 @@ public class MaintenanceController : ControllerBase
 
         if (this._serviceOptions.Value.InMaintenance)
         {
-            result = new MigrationResult();
+            return this.Ok(new MigrationResult());
         }
 
-        return this.Ok(result);
+        return this.Ok();
     }
 }
--- a/webapi/Skills/ChatSkills/SemanticChatMemoryExtractor.cs
+++ b/webapi/Skills/ChatSkills/SemanticChatMemoryExtractor.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Threading;
@@ -140,11 +141,11 @@ internal static class SemanticChatMemoryExtractor
     {
         var memoryCollectionName = SemanticChatMemoryExtractor.MemoryCollectionName(chatId, memoryName);
 
-#pragma warning disable CA1031 // Each connector may throw different exception type
+        IList<MemoryQueryResult>? memories = null;
         try
         {
             // Search if there is already a memory item that has a high similarity score with the new item.
-            var memories = await semanticTextMemory.SearchAsync(
+            memories = await semanticTextMemory.SearchAsync(
                     collection: memoryCollectionName,
                     query: item.ToFormattedString(),
                     limit: 1,
@@ -153,8 +154,16 @@ internal static class SemanticChatMemoryExtractor
                 )
                 .ToListAsync()
                 .ConfigureAwait(false);
+        }
+        catch (Exception connectorException) when (!connectorException.IsCriticalException())
+        {
+            // A store exception might be thrown if the collection does not exist, depending on the memory store connector.
+            logger.LogError(connectorException, "Cannot search collection {0}", memoryCollectionName);
+        }
 
-            if (memories.Count == 0)
+        try
+        {
+            if ((memories?.Count ?? 0) == 0)
             {
                 await semanticTextMemory.SaveInformationAsync(
                     collection: memoryCollectionName,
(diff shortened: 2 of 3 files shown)


webapi/Controllers/MaintenanceController.cs
```diff
@@ -55,9 +55,9 @@ public class MaintenanceController : ControllerBase
 
         if (this._serviceOptions.Value.InMaintenance)
         {
-            result = new MigrationResult();
+            return this.Ok(new MigrationResult());
```
dehoward, line 58: the variable above can be removed now

webapp/src/components/views/BackendProbe.tsx
```diff
@@ -44,14 +44,17 @@ export const BackendProbe: FC<IData> = ({ uri, onBackendFound }) => {
...
-                if (json === null) {
-                    dispatch(setMaintenance(false));
-                    onBackendFound();
-                }
-
-                model.current = json as IMaintenance | null;
+                result
+                    .json()
+                    .then((data) => {
+                        // Has payload is maintenance
```
dehoward, line 50: nit: comment is hard to undertand gramatically

alliscode (approved):

webapi/Skills/ChatSkills/SemanticChatMemoryExtractor.cs
```diff
@@ -165,12 +174,11 @@ internal static class SemanticChatMemoryExtractor
                 );
             }
         }
-        catch (Exception connectorException)
+        catch (Exception connectorException) when (!connectorException.IsCriticalException())
```
teresaqhoang, line 177: Do we need try/catch this here? The exception should already be thrown when it's searched above, right?

If not, I would suggest moving this into the try catch block above as well to reduce duped code.

```
if ((memories?.Count ?? 0) == 0)
            {
                await semanticTextMemory.SaveInformationAsync(
                    collection: memoryCollectionName,
@@ -165,12 +174,11 @@ internal static class SemanticChatMemoryExtractor
                );
            }
```

webapp/src/components/views/BackendProbe.tsx
```diff
@@ -44,14 +44,17 @@ export const BackendProbe: FC<IData> = ({ uri, onBackendFound }) => {
...
+                        setModel(data as IMaintenance);
+                    })
+                    .catch(() => {
```
teresaqhoang, line 53: This catch block is meant catch exceptions. If the API errors, it doesn't necessarily mean that the backend is not in maintenance mode and it would prematurely bail out of this health check. 

From the code in `MaintenanceController` above, it seems like the only condition where the backend is not in maintenance mode is if the API returns a 200 OK with no body, so shouldn't this bebe in the then block above? Under a conditional that checks for any empty body
```
if (!data)
{
 // No maintenance payload
                        dispatch(setMaintenance(false));
                        onBackendFound();
}
```

webapi/Skills/ChatSkills/SemanticChatMemoryExtractor.cs
crickman: [Connector](https://github.com/microsoft/chat-copilot/pull/365)
https://github.com/microsoft/chat-copilot/issues/351
https://github.com/microsoft/chat-copilot/issues/215

webapp/src/components/views/BackendProbe.tsx
crickman: The exception is the JSON parsing of undefined body.  Since the body is a stream, its a bit indirect to evaluate up front...hence, the response to the error.  I believe the API has returned response already at this point.

webapp/src/components/views/BackendProbe.tsx
crickman: Updated

webapi/Controllers/MaintenanceController.cs
crickman: Updated

webapp/src/components/views/BackendProbe.tsx
teresaqhoang: Ah I see now, can you add a comment to clarify the JSON exception we're catching here?

webapi/Skills/ChatSkills/SemanticChatMemoryExtractor.cs
teresaqhoang: Yes, but since `semanticTextMemory.SearchAsync` is already called above, the missing index exception should already be thrown on that call

I also think you can consolidate both calls into the same try catch block 
image

teresaqhoang: Never mind, I see the first one can throw a 404 if there's no related memories

webapi/Skills/ChatSkills/SemanticChatMemoryExtractor.cs
```diff
@@ -153,8 +154,16 @@ internal static class SemanticChatMemoryExtractor
                 )
                 .ToListAsync()
                 .ConfigureAwait(false);
+        }
+        catch (Exception connectorException) when (!connectorException.IsCriticalException())
+        {
+            // A store exception might be thrown if the collection does not exist, depending on the memory store connector.
```
teresaqhoang, line 160:
```suggestion
            // A store exception might be thrown if the collection does not exist or there are no related memories, depending on the memory store connector.
```

teresaqhoang (approved):

PR #367, opened 2023-09-19
Order change (include model response)

crickman:
### Contribution Checklist


momuno:
Hey @TaoChenOSU.  When I looked at the code, it seems like the "memory" is being extracted from the chat history and stored in the semantic-memory before the chat history is updated with the most recent completion (aka chatMessage).  Do you know if there was a design reason for this?  Thanks. (And thank you to @crickman for creating the PR for me.)

TaoChenOSU:
> Hey @TaoChenOSU. When I looked at the code, it seems like the "memory" is being extracted from the chat history and stored in the semantic-memory before the chat history is updated with the most recent completion (aka chatMessage). Do you know if there was a design reason for this? Thanks. (And thank you to @crickman for creating the PR for me.)

Hey @momuno, you're right. Good catch! I believe this is not intentional. We should move the memory extraction to after the chatMessage is saved. However, we should not move the memory extraction to before the token calculation since memory extraction consume tokens as well.

PR #377, opened 2023-09-22
Modify deployment scripts to serve static files from web service

glahaye:
### Motivation and Context
This change enables having a generic single package containing both backend and frontend that can be deployed using the "Deploy to Azure" button.

It also reduces the complexity and cost of deployments.

### Description
- Remove usage of SWA and its CLI.
- Make the backend serve files.
- Build static files with yarn and copy them to package deployed to Azure.
- Introduced a middleware that inserts the configuration values needed by the frontend in its source code.
- Modified deployment scripts accordingly

### Contribution Checklist


scripts/deploy/deploy-webapi.ps1
```diff
@@ -42,22 +42,55 @@ if ($LASTEXITCODE -ne 0) {
...
     Write-Error "Could not get Azure WebApp resource name from deployment output."
     exit 1
 }
 
-Write-Host "Azure WebApp name: $webappName"
+Write-Host "Azure WebApp name: $webApiName"
 
 Write-Host "Configuring Azure WebApp to run from package..."
-az webapp config appsettings set --resource-group $ResourceGroupName --name $webappName --settings WEBSITE_RUN_FROM_PACKAGE="1" | out-null
+az webapp config appsettings set --resource-group $ResourceGroupName --name $webApiName --settings WEBSITE_RUN_FROM_PACKAGE="1" | out-null
```
alliscode, line 57: Please add `-o none` to this command so that it does not output anything. This command has the potential to leaks secrets otherwise.

alliscode: If you see this command anywhere else, please add it there also.

scripts/deploy/deploy-webapi.ps1
```diff
@@ -42,22 +42,55 @@ if ($LASTEXITCODE -ne 0) {
...
+        --method PATCH `
+        --uri "https://graph.microsoft.com/v1.0/applications/$objectId" `
+        --headers 'Content-Type=application/json' `
+        --body $body
+}
+if ($LASTEXITCODE -ne 0) {
+    exit $LASTEXITCODE
+}
+
+Write-Host "To verify your deployment, go to 'https://$webApiUrl/' in your browser."
```
alliscode, line 96: I think this needs to be wrapped in an condition that can be opted into because setting the redirectUri on the app requires very high privileges on the subscription that most time will not be present for this script. I don't think this works with our internal hosted environements for example.

scripts/deploy/deploy-webapi.ps1
```diff
@@ -42,22 +42,55 @@ if ($LASTEXITCODE -ne 0) {
...
     exit $LASTEXITCODE
-}
\ No newline at end of file
+}
+
+
+$origin = "https://$webApiUrl"
+Write-Host "Ensuring '$origin' is included in AAD app registration's redirect URIs..."
+
+$webapiSettings = $(az webapp config appsettings list --name $webapiName --resource-group $ResourceGroupName | ConvertFrom-JSON)
```
alliscode, line 72: Can you please confirm that this command only returns the name in the output. The webApi settings contains secrets and the output can end up in the github logs.

scripts/deploy/deploy-webapi.sh
```diff
@@ -70,28 +70,54 @@ fi
...
     echo "Could not get Azure WebApp resource name from deployment output."
     exit 1
 fi
 
-echo "Azure WebApp name: $WEB_APP_NAME"
+echo "Azure WebApp name: $WEB_API_NAME"
 
 echo "Configuring Azure WebApp to run from package..."
-az webapp config appsettings set --resource-group $RESOURCE_GROUP --name $WEB_APP_NAME --settings WEBSITE_RUN_FROM_PACKAGE="1"
+az webapp config appsettings set --resource-group $RESOURCE_GROUP --name $WEB_API_NAME --settings WEBSITE_RUN_FROM_PACKAGE="1"
```
alliscode, line 87: Same comments for this file as in the ps1.

scripts/deploy/main.bicep
```diff
@@ -248,10 +248,6 @@ resource appServiceWebConfig 'Microsoft.Web/sites/config@2022-09-01' = {
         name: 'MemoryStore:Qdrant:Port'
         value: '443'
       }
-      {
-        name: 'MemoryStore:AzureCognitiveSearch:UseVectorSearch'
-        value: 'true'
-      }
```
alliscode, line 254: Why is this being deleted?

scripts/deploy/main.bicep
```diff
@@ -304,10 +308,6 @@ resource appServiceWebConfig 'Microsoft.Web/sites/config@2022-09-01' = {
         name: 'Logging:ApplicationInsights:LogLevel:Default'
         value: 'Warning'
       }
-      {
-        name: 'ApplicationInsights:ConnectionString'
-        value: appInsightsWeb.properties.ConnectionString
-      }
```
alliscode, line 310: Is this not being used?

webapi/Middleware/ResponseTokenReplacerMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    private string ReplaceTokens(HttpContext context, string text)
+    {
+        IConfiguration appSettings = context.RequestServices.GetRequiredService<IConfiguration>();
+
+        MatchCollection matches = this._tokenRegex.Matches(text);
+
+        foreach (string token in matches.Select(m => m.Groups[1].Value).Distinct())
+        {
+            string? settingValue = appSettings.GetValue<string>(token);
```
alliscode, line 87: Where does this token come from?

scripts/deploy/deploy-webapi.ps1
glahaye: OK.  Replacing the redirections of output to null with -o none

scripts/deploy/main.bicep
glahaye: This has been removed from code ages ago but was forgotten in the bicep file

scripts/deploy/main.bicep
glahaye: No.  We use the standard APPLICATIONINSIGHTS_CONNECTION_STRING instead

webapi/Middleware/ResponseTokenReplacerMiddleware.cs
glahaye: This is whatever is between <-=token=-></-=token=-> tags in the processed file

scripts/deploy/deploy-webapi.ps1
glahaye: This code already existed and ran fine in deploy-webapp.  But, sure, I can totally put a switch to optionally skip this step.

scripts/deploy/deploy-webapi.ps1
glahaye: The command returns EVERYTHING in webapiSettings but doesn't print anything on the console.

Is that a problem?

crickman: Is this two changes in one PR:

1. The deployment update
2. The middleware/options update

Or are they connected?

glahaye: The change consists of distributing CC as one package instead of two (one for the backend and another for the frontend).

This means clicking on "Deploy to Azure" is all you'll need to do to have a complete working deployment (no more additional shell scripts needed).

scripts/deploy/deploy-webapi.ps1
```diff
@@ -74,25 +91,68 @@ if ($DeploymentSlot) {
...
+    }
+}
+
+if ($RegisterPluginCors) {
+    foreach ($pluginName in $pluginNames) {
+        $allowedOrigins = $((az webapp cors show --name $pluginName --resource-group $ResourceGroupName --subscription $Subscription | ConvertFrom-Json).allowedOrigins)
+        foreach ($address in $origins) {
+            $origin = "https://$address"
+            Write-Host "Ensuring '$origin' is included in CORS origins for plugin '$pluginName'..."
+            if (-not $allowedOrigins -contains $origin)) {
```
dehoward, line 151: syntax error :/
```suggestion
            if (-not $allowedOrigins -contains $origin) {
```

scripts/deploy/deploy-webapi.sh
```diff
@@ -36,6 +38,11 @@ while [[ $# -gt 0 ]]; do
         ;;
         -p|--package)
         PACKAGE_FILE_PATH="$2"
+        shiftregister-app
+        shift
+        ;;
+        -r|--package)
```
dehoward, line 44:
```suggestion
        -r|--register-app)
```

scripts/deploy/deploy-webapi.sh
```diff
@@ -36,6 +38,11 @@ while [[ $# -gt 0 ]]; do
         ;;
         -p|--package)
         PACKAGE_FILE_PATH="$2"
+        shiftregister-app
```
dehoward, line 41: is this right??
```suggestion
        shift
```

scripts/deploy/package-webapi.sh
```diff
@@ -100,6 +107,41 @@ if [ $? -ne 0 ]; then
...
+if [[ -z "$NO_FRONTEND" ]]; then
+    echo "Building static frontend files..."
+
+    ENV_FILE_PATH="$SCRIPT_ROOT/../../webapp/.env"
+    echo "Writing environment variables to '$ENV_FILE_PATH'..."
+    echo "REACT_APP_BACKEND_URI=<-=TOKEN=->Frontend:BackendUri</-=TOKEN=->" > $ENV_FILE_PATH
+    echo "REACT_APP_AUTH_TYPE=AzureAd" >> $ENV_FILE_PATH
+    echo "REACT_APP_AAD_AUTHORITY=<-=TOKEN=->Authentication:AzureAd:Instance</-=TOKEN=->/<-=TOKEN=->Authentication:AzureAd:TenantId</-=TOKEN=->" >> $ENV_FILE_PATH
+    echo "REACT_APP_AAD_CLIENT_ID=<-=TOKEN=->Frontend:AadClientId</-=TOKEN=->" >> $ENV_FILE_PATH
+    echo "REACT_APP_AAD_API_SCOPE=api://<-=TOKEN=->Authentication:AzureAd:ClientId</-=TOKEN=->/<-=TOKEN=->Authentication:AzureAd:Scopes</-=TOKEN=->" >> $ENV_FILE_PATH
```
dehoward, line 119: as you know, we can remove this now!

dehoward (changes requested):

scripts/deploy/deploy-webapi.ps1
glahaye: Yes. I've merged a few things in a gung ho manner last night.  Now I'm testing the results and found this bug as well.  Fixed.

scripts/deploy/deploy-webapi.sh
glahaye: That was a copy/paste error.  Fixed thanks.

scripts/deploy/package-webapi.sh
glahaye: Yep.  Done on a local copy that I'm testing right now.

scripts/deploy/deploy-webapi.sh
glahaye: ???

scripts/deploy/package-webapi.ps1
```diff
@@ -22,13 +22,19 @@ param(
...
+    $Version = "0.0.0",
 
     [string]
     # Additional information given in version info.
-    $InformationalVersion = ""
+    $InformationalVersion = "",
+    
+    [bool]
+    # Whether to build frontend files (true by default)
+    $BuildFrontendFiles = $true
```
dehoward, line 33: it seems like we should align this flag with the one in `.sh`?

glahaye: Done

dehoward (approved):

raj-microsoft:
the docs here could be updated to reflect the new deployment steps [sk-frontend-deployment](https://learn.microsoft.com/en-us/semantic-kernel/chat-copilot/deploy-to-azure#deploy-the-chat-copilot-frontend)

Also is it required to create 2 separate app registrations, now that both frontend & backend is served from within one appservice

glahaye:
> the docs here could be updated to reflect the new deployment steps [sk-frontend-deployment](https://learn.microsoft.com/en-us/semantic-kernel/chat-copilot/deploy-to-azure#deploy-the-chat-copilot-frontend)
> 
> Also is it required to create 2 separate app registrations, now that both frontend & backend is served from within one appservice

@raj-microsoft You still need 2 app registrations. The frontend files can be hosted anywhere independently from the backend. They are now hosted by the backend by default to reduce the number of moving parts and to reduce the number of steps required to get to a functional Chat Copilot deployment.

Thank you for pointing out the disconnect with the documentation on learn.microsoft.com.  I will go and fix that.

PR #379, opened 2023-09-22
Fix issue 378 to able to use scripts with spaces in full path

MaggieWang0:
### Description

Update the command to handle the spaces in path of backend script

### Contribution Checklist

### DIFF
--- a/scripts/Start.ps1
+++ b/scripts/Start.ps1
@@ -3,11 +3,10 @@
 Initializes and runs both the backend and frontend for Chat Copilot.
 #>
 
-$BackendScript = Join-Path "$PSScriptRoot" 'Start-Backend.ps1'
 $FrontendScript = Join-Path "$PSScriptRoot" 'Start-Frontend.ps1'
 
 # Start backend (in new PS process)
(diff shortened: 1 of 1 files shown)


dehoward (dismissed):

gitri-ms:
We explicitly chose to use `pwsh` in these scripts and not `powershell` because `pwsh` is supported cross-platform. We are also utilizing some functionality that is only supported by version 6 and above. I understand the desire to not install an extra thing, but ultimately the pros outweighed the con in that decision. We previously had instructions in our README that users need to [install PowerShell Core](https://github.com/PowerShell/PowerShell) but it looks like that has gotten removed. I will open an item to add that instruction back.

dehoward:
thanks, @gitri-ms! 

@MaggieWang0 I will go ahead and close this PR. thanks for raising the issue!

MaggieWang0:
Hi Gina,
Thanks for the explanation. I sent the PR mainly for the space issue and I didn’t notice the difference between powershell and pwsh. Solving the space issue is enough to me.

Thanks,
Maggie

Issue #378, opened 2023-09-22
Not able to start backend with spaces in script path


MaggieWang0:
**Describe the bug**
Not able to start the Start-Backend.ps1 when I put the repo in a path with spaces.

**To Reproduce**
1. Put the repo in a path with spaces
2. run '.\Start.ps1' as Readme
3. error: the backend won't be able to start, and the window exist due to no '-NoExit'

**Expected behavior**
Start backend

**Screenshots**
<img width="359" alt="image" src="https://github.com/microsoft/chat-copilot/assets/28853205/56367a65-9958-4a1d-8cfa-947fde3c9cb4">

**Platform**
 - OS: Windows
 - IDE: Powershell
 - Language: ps1
 - Source: main branch of repository

**Additional context**

samleegithub:
I ran into the same problem. Here is a fix to line 10 in Start.ps1:

```
Start-Process pwsh -ArgumentList "-noexit `"$BackendScript`""
```

dehoward:
fixed by #385

PR #383, opened 2023-09-22
Update Start.ps1

samleegithub:
### Motivation and Context

Fixes this issue: https://github.com/microsoft/chat-copilot/issues/378

### Description

- Removed invalid "-command" argument.
- Added "-noexit" argument to prevent the new pwsh window from closing if there is an error. This helps with debugging if there is an issue.
- Added escaped double-quotes around the backend script path so that it does not fail when there is a space in the path to the script.

### Contribution Checklist


dehoward:
hi @samleegithub, thanks for this contribution!

there are currently 2 other PRs out with fixes to the `Start.ps1` script (#383 & #385), both of which also address the usage of the `pwsh` command, which doesn't work on all machines. I will close your PR in favor of getting one of these merged, since I think it is a good time to fix the command as well.

PR #385, opened 2023-09-22
Update scripts\Start.ps1

crickman:
### Motivation and Context

1. Explicity detect that powershell "core" is installed (pwsh vs powershell)
2. Execute `Start-Backend.ps1` even if path has spaces

### Contribution Checklist

### DIFF
--- a/scripts/Start.ps1
+++ b/scripts/Start.ps1
@@ -3,11 +3,21 @@
 Initializes and runs both the backend and frontend for Chat Copilot.
 #>
 
+# Verify "Core" version of powershell installed (not "Desktop"): https://aka.ms/powershell
+$ErrorActionPreference = 'Ignore'
+$cmd = get-command 'pwsh'
+$ErrorActionPreference = 'Continue'
+
+if (!$cmd) {
+	Write-Warning "Please update your powershell installation: https://aka.ms/powershell"
+	return;
+}
+
 $BackendScript = Join-Path "$PSScriptRoot" 'Start-Backend.ps1'
 $FrontendScript = Join-Path "$PSScriptRoot" 'Start-Frontend.ps1'
 
 # Start backend (in new PS process)
-Start-Process pwsh -ArgumentList "-command $BackendScript"
+Start-Process pwsh -ArgumentList "-command ""& '$BackendScript'"""
 # check if the backend is running before proceeding
 $backendRunning = $false
 
--- a/scripts/Start.ps1
+++ b/scripts/Start.ps1
@@ -9,8 +9,8 @@ $cmd = get-command 'pwsh'
 $ErrorActionPreference = 'Continue'
 
 if (!$cmd) {
-	Write-Warning "Please update your powershell installation: https://aka.ms/powershell"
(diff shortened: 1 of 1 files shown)


scripts/Start.ps1
```diff
@@ -3,11 +3,21 @@
 Initializes and runs both the backend and frontend for Chat Copilot.
 #>
 
+# Verify "Core" version of powershell installed (not "Desktop"): https://aka.ms/powershell
```
dehoward, line 6: do you know why we're using `pwsh` instead of `powershell` if it requires an extra step on Windows to work? I see that this changed last month with the security upates, so maybe @gitri-ms knows as well. 

given that we already have a `start.sh` script for bash, it seems like using the out-of-the-box Windows Powershell command would be fine

glahaye (approved):

scripts/Start.ps1
dehoward: there is another PR out (#379) that changes the script back to use `powershell`, since I'm guessing that user had issues with `pwsh` as well. imo this is the way to go since it doesn't require any extra steps on Windows, but I'm curious to know why switched in the first place

crickman: This isn't a new requirement (in terms of specifying "pwsh" in the script).  The main idea here is that in the current implementation an confusing error would result.  I'm trying to avoid analyzing our script execution stack; although, you raise a great point...there's likely an opportunity for deeper hygiene here.

I had to install "powershell core" as part of my SK onboarding so I was under the impression that is was just common.

crickman: Since Gina is out today, maybe take this band-aid and then determine if we can back away from "pwsh" next week when she's back.

samleegithub: If powershell works, I would vote for changing it back to powershell instead of pwsh.

I had to install Powershell 7 because it was calling pwsh and it took me a while to figure out windows powershell and powershell 7 were different. This wasn't called out in the readme and I was thinking about adding powershell 7 as a requirement.

dehoward: > Since Gina is out today, maybe take this band-aid and then determine if we can back away from "pwsh" next week when she's back.

@crickman yeah I like that idea. we can keep the other PR open until next week but get this one in today.

dehoward (approved):

scripts/Start.ps1
gitri-ms: We explicitly chose to use `pwsh` in these scripts and not `powershell` because `pwsh` is supported cross-platform. We are also utilizing some functionality that is only supported by version 6 and above. I understand the desire to not install an extra thing, but ultimately the pros outweighed the con in that decision. We previously had instructions in our README that users need to [install PowerShell Core](https://github.com/PowerShell/PowerShell) but it looks like that has gotten removed. I can open an item to add that back.

glahaye: Do we care about the cross-platform aspect?  Why do we have equivalent shell scripts to all our PowerShell ones, then?

PR #409, opened 2023-09-27
🚀 Feature: Re-run saved plans

teresaqhoang:
### Motivation and Context

This PR adds support for re-running saved plans previously generated by the model. Once re-run, these plans are considered `derived` plans. This PR also includes a refactor of the webapi, specifically `ChatSkill` and `ChatController`, to separate plan execution from `/chat` requests into an isolated `/processplan` request flow. 

### Description

webapi
- The `ExecutePlanParameters` class has been added to the webapi/Models/Request directory as request parameters for `/processplan` request, which is handled by the addition of a new ProcessPlan function to the ChatController.
- `ChatSkill `has been updated to handle plan execution and re-execution and can also support derived plans. 
- Added logic for checking that plan is actionable by supported planners
- Set planner supplement in prompt options
- ChatHistory and Relevant memories is now omitted if plan yields results, to help prioritize relevant data.
- Refactored flow after meta prompt completion so Chat history is updated before semantic memory is generated. 

webapp
- The `PlansTab` component now displays the user intent as the goal of each plan instead of the input and includes a new `PlanDialogView` component for displaying plan details.  The `PlanDialogView` component includes a formatted view of the plan, as well as a JSON view for developers.
- The `PlanViewer` component has been updated to remove the `getResponse` prop and improve its functionality and readability.
- Added the definition of a `Plan` interface and a `ProposedPlan` interface
- The `useChat` hook has been updated to support plan processing, as well as changes related to Plans and Plugins, including updates to the `PluginConnector` and `PluginWizard` components to use new styles.

new tabbed PlanDialogView in PlansTab
image

Confirmation Dialog if user wants to re-run plan
image

If user clicks "Yes, proceed", then they will automatically be redirected back to chat window and two new messages will be generated: one signaling user action and a bot message to save the derived plan. Plan execution will start immediately.
image

Full flow
Rerun Plans Flow

### Contribution Checklist


glahaye (dismissed):

webapi/Models/Storage/ChatMessage.cs
```diff
@@ -259,4 +259,12 @@ private static bool IsPlan(string response)
         var planPrefix = "proposedPlan\":";
         return response.IndexOf(planPrefix, StringComparison.Ordinal) != -1;
     }
+
+    /// <summary>
+    /// Convert a chat message to a generic string.
+    /// </summary>
+    /// <param name="role"></param>
+    /// <param name="content"></param>
+    /// <returns></returns>
```
TaoChenOSU, line 268: missing xml doc string for return and params

webapi/Models/Storage/ChatMessage.cs
```diff
@@ -259,4 +259,12 @@ private static bool IsPlan(string response)
...
+    public static string ToString(AuthorRoles role, string content) => $"{role}: {content}";
```
TaoChenOSU, line 269: Does it make sense to create a static function here especially the class itself already has a `ToString` method? We can have a util module. Same for the `SafeInvokeAsync` method. `SemanticKernelExtensions` is probably not the best place to put them in.

glahaye: Oh, I hadn't even noticed this was a static method!

webapi/Skills/ChatSkills/ExternalInformationSkill.cs
```diff
@@ -94,7 +98,7 @@ public class ExternalInformationSkill
             return string.Empty;
         }
 
-        var contextString = string.Join("\n", context.Variables.Select(v => $"{v.Key}: {v.Value}"));
+        var contextString = string.Join("\n", context.Variables.Where(v => !(v.Key.Contains("TokenUsage", StringComparison.CurrentCultureIgnoreCase) || v.Key.Contains("tokenLimit", StringComparison.CurrentCultureIgnoreCase) || v.Key.Contains("chatId", StringComparison.CurrentCultureIgnoreCase))).Select(v => $"{v.Key}: {v.Value}"));
```
TaoChenOSU, line 101: nit: maybe worth putting this in a method with better description of what this does and why this has to be done

webapi/Models/Storage/ChatMessage.cs
teresaqhoang: Moved this out to `PromptUtils.FormatChatHistoryMessage` and moved the `SafeInvokeAsync` methods to `AsyncUtils`

webapi/Skills/ChatSkills/ExternalInformationSkill.cs
teresaqhoang: Good call, moved this out to new ` private string GetChatContextString(SKContext context)` method

webapi/Models/Storage/ChatMessage.cs
teresaqhoang: Done in `PromptUtils.FormatChatHistoryMessage`

glahaye (approved):

TaoChenOSU (dismissed):

webapi/Skills/ChatSkills/ChatSkill.cs
```diff
@@ -337,43 +336,165 @@ public async Task<string> ExtractAudienceAsync(SKContext context, CancellationTo
...
+        // Set the system description in the prompt options.
+        await this.SetSystemDescriptionAsync(chatId, cancellationToken);
+
+        // Clone the context to avoid modifying the original context variables.
+        var chatContext = context.Clone();
+        chatContext.Variables.Set("knowledgeCutoff", this._promptOptions.KnowledgeCutoffDate);
+
+        // Save this new message to memory such that subsequent chat responses can use it
+        await this.UpdateBotResponseStatusOnClientAsync(chatId, "Saving user message to chat history", cancellationToken);
+        var newUserMessage = await this.SaveNewMessageAsync(message, userId, userName, chatId, ChatMessage.ChatMessageType.Message.ToString(), cancellationToken);
```
dehoward, line 393: looks like we could have a helper fn for this logic and use it in both routes?

dehoward (approved):

webapi/Skills/ChatSkills/ChatSkill.cs
teresaqhoang: Can leave for another PR if we want to refactor it, but I figured I'd leave it like this for now because it allows more flexibility in manipulating the chatContext as well as fine tuning the system description, if needed.

TaoChenOSU (approved):

Issue #388, opened 2023-09-23
Bug: if I modified the configuration of `Prompts.WorkingMemoryName` and `LongTermMemoryName`, you should use them as Key for access `MemoryMap` in `ChatMemoryController`


AwesomeYuer:
**Describe the bug**
if I modified the configuration of `Prompts.WorkingMemoryName` and `LongTermMemoryName`, you should use them for access in `ChatMemoryController`

**To Reproduce**
1. modified the configuration of `Prompts.WorkingMemoryName` and `LongTermMemoryName`
```json
"Prompts": {
  "WorkingMemoryName": "W",
  "LongTermMemoryName": "L"
}
```
2. access `persona` tab on frontend, will occurs 400 error, that is access the url as below:

`https://localhost:40443/chatMemory/wxuse_2306_b17893d883b34657a462841a8208c938/WorkingMemory`

`https://localhost:40443/chatMemory/wxuse_2306_b17893d883b34657a462841a8208c938/LongTermMemory`

because you didn't use the configuration as above

**Expected behavior**
no error!

**Screenshots**
Fix suggestions

![image](https://github.com/microsoft/chat-copilot/assets/1026479/fd499676-cc66-40ab-ae10-be07c0b8b803)

![image](https://github.com/microsoft/chat-copilot/assets/1026479/38046e7c-2d64-4a3d-93d3-099f757b10a1)

**Platform**
 - OS: [e.g. Windows, Mac]
 - IDE: [e.g. Visual Studio, VS Code]
 - Language: [e.g. C#, Python]
 - Source: [e.g. NuGet package version 0.1.0, pip package version 0.1.0, main branch of repository]

**Additional context**

AwesomeYuer:
found another impact and Fix suggestions:

![image](https://github.com/microsoft/chat-copilot/assets/1026479/c0bc22c4-227b-4bc1-bafe-d5f8e00a2a38)

![image](https://github.com/microsoft/chat-copilot/assets/1026479/9dd089b4-eae7-40b4-98d2-bac09bd622aa)

crickman:
Thank you for reporting this.  I'm familiar with the code/config to which you are referring and and will review.

crickman:
@AwesomeYuer - Curious, do you find it useful that these keys can be configured?  If so, why?

AwesomeYuer:
Sure!
Since I modified your code, changed `chatId` to a `string` from `Guid`, 
[...]

PR #410, opened 2023-09-27
Using semantic memory type enum instead of string in API contract.

alliscode:
### Motivation and Context

Currently the memory retrieval API requires the name of the memory as input, however, the memoryName is something that's configured in the backend and is not known by external clients. This PR refactors things to use a memoryType based on a stable enum as the public interface to the API which allows the memoryName to remain a configuration detail of the backend.   
-->

See issue #388 

### Contribution Checklist

### DIFF
--- a/webapi/Controllers/ChatMemoryController.cs
+++ b/webapi/Controllers/ChatMemoryController.cs
@@ -51,7 +51,7 @@ public class ChatMemoryController : ControllerBase
     /// </summary>
     /// <param name="semanticTextMemory">The semantic text memory instance.</param>
     /// <param name="chatId">The chat id.</param>
-    /// <param name="memoryName">Name of the memory type.</param>
+    /// <param name="memoryType">Type of the memory type. Must map to a member of <see cref="SemanticMemoryType"/>.</param>
     [HttpGet]
     [Route("chatMemory/{chatId:guid}/{memoryType}")]
     [ProducesResponseType(StatusCodes.Status200OK)]
@@ -66,7 +66,8 @@ public class ChatMemoryController : ControllerBase
         // https://github.com/microsoft/chat-copilot/security/code-scanning/1
         var sanitizedChatId = GetSanitizedParameter(chatId);
 
-        if (!this._promptOptions.TryGetMemoryName(memoryType, out string memoryName))
+        // Map the requested memorType to the memory store container name
+        if (!this._promptOptions.TryGetMemoryContainerName(memoryType, out string memoryConatinerName))
         {
             this._logger.LogWarning("Memory type: {0} is invalid.", memoryType);
             return this.BadRequest($"Memory type: {memoryType} is invalid.");
@@ -88,7 +89,7 @@ public class ChatMemoryController : ControllerBase
             // Search if there is already a memory item that has a high similarity score with the new item.
             var filter = new MemoryFilter();
             filter.ByTag("chatid", chatId);
-            filter.ByTag("memory", memoryName);
+            filter.ByTag("memory", memoryConatinerName);
             filter.MinRelevance = 0;
 
             var searchResult =
@@ -98,7 +99,7 @@ public class ChatMemoryController : ControllerBase
                     relevanceThreshold: 0,
                     resultCount: 1,
                     chatId,
-                    memoryName)
+                    memoryConatinerName)
                 .ConfigureAwait(false);
 
             foreach (var memory in searchResult.Results.SelectMany(c => c.Partitions))
@@ -109,7 +110,7 @@ public class ChatMemoryController : ControllerBase
         catch (Exception connectorException) when (!connectorException.IsCriticalException())
         {
             // A store exception might be thrown if the collection does not exist, depending on the memory store connector.
-            this._logger.LogError(connectorException, "Cannot search collection {0}", memoryName);
+            this._logger.LogError(connectorException, "Cannot search collection {0}", memoryConatinerName);
         }
 
         return this.Ok(memories);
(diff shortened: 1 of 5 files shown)


webapi/Controllers/ChatMemoryController.cs
```diff
@@ -52,19 +53,24 @@ public class ChatMemoryController : ControllerBase
     /// <param name="chatId">The chat id.</param>
     /// <param name="memoryName">Name of the memory type.</param>
```
teresaqhoang, line 54:
```suggestion
    /// <param name="memoryType">Type of the memory.</param>
```

webapi/Controllers/ChatMemoryController.cs
```diff
@@ -52,19 +53,24 @@ public class ChatMemoryController : ControllerBase
...
         [FromRoute] string chatId,
-        [FromRoute] string memoryName)
+        [FromRoute] string memoryType)
     {
         // Sanitize the log input by removing new line characters.
         // https://github.com/microsoft/chat-copilot/security/code-scanning/1
         var sanitizedChatId = GetSanitizedParameter(chatId);
-        var sanitizedMemoryName = GetSanitizedParameter(memoryName);
+
+        if (!this._promptOptions.TryGetMemoryName(memoryType, out string memoryName))
```
teresaqhoang, line 69: What's the difference between memoryType and memoryName?

webapi/Options/PromptsOptions.cs
```diff
@@ -161,4 +163,29 @@ public class PromptsOptions
...
     /// <returns>A shallow copy of the options.</returns>
     internal PromptsOptions Copy() => (PromptsOptions)this.MemberwiseClone();
+
+    /// <summary>
+    /// Tries to retrieve the memoryName asociated with the specified memory type.
+    /// </summary>
+    internal bool TryGetMemoryName(string memoryType, out string memoryName)
+    {
+        memoryName = "";
+        if (!Enum.TryParse<SemanticMemoryType>(memoryType, true, out SemanticMemoryType semanitcMemoryType))
```
teresaqhoang, line 173:
```suggestion
        if (!Enum.TryParse<SemanticMemoryType>(memoryType, true, out SemanticMemoryType semanticMemoryType))
```

webapi/Options/PromptsOptions.cs
```diff
@@ -161,4 +163,29 @@ public class PromptsOptions
     /// </summary>
     /// <returns>A shallow copy of the options.</returns>
     internal PromptsOptions Copy() => (PromptsOptions)this.MemberwiseClone();
+
+    /// <summary>
+    /// Tries to retrieve the memoryName asociated with the specified memory type.
+    /// </summary>
+    internal bool TryGetMemoryName(string memoryType, out string memoryName)
```
glahaye, line 170: You probably want to name this TryGetMemoryType() ?

webapi/Controllers/ChatMemoryController.cs
```diff
@@ -89,7 +88,7 @@ public class ChatMemoryController : ControllerBase
             // Search if there is already a memory item that has a high similarity score with the new item.
             var filter = new MemoryFilter();
             filter.ByTag("chatid", chatId);
-            filter.ByTag("memory", sanitizedMemoryName);
+            filter.ByTag("memory", memoryName);
```
glahaye, line 91: Although probably out of the scope of this change, you might want to rename all instances of "memoryName" in all of our code to "memoryType".  If you don't do it, I will at one point.  This has been annoying me for a while.  Feel free to ignore for now.  As I said, I'll get around to it if you don't.

webapi/Controllers/ChatMemoryController.cs
alliscode: MemoryType maps to an enum, and the values of that are stable. Memory name is configurable in appsettings.json and maps to the name of the index or tag used in the memory store. o the idea is that the API contract uses memoryType because it won't have it's values changed, and then the memoryName can be changed in appsettings.json without breaking anything.

webapi/Controllers/ChatMemoryController.cs
alliscode: I'll go look for places that need to be renamed. But keep in mind that there are still uses of memoryName when it comes to accessing memory - memoryType is just the public contract for the API so that things don't break when the memory name in appsettings.json is changed.

webapi/Options/PromptsOptions.cs
alliscode: No actually this method is using the memoryType to get the memoryName. Think about it this way, we define two types of memory, LongTermMemory, and WorkingMemory. The actual names of the indexes or tags that the memories store uses for those memory types are configurable and therfore can be changed, so we need a way to map from the typo of memory asked for in the API, to the name of the index or tag where it's stored in the memory store..

glahaye: So, really, it's GetMemoryTypeName()?

webapi/Controllers/ChatMemoryController.cs
glahaye: I guess what I'm saying is that the type of memory we use should never have been named MemoryName but rather MemoryType.  It's confusing.  This PR is a case in point,

alliscode: Yes I agree, it's super confusing. I'm renaming some of it to `memoryContainerName` to try to clarify that it's related to a database container name rather than the type of memory.

webapi/Options/PromptsOptions.cs
alliscode: You're right, that would be more correct, although maybe more confusing at this point. I'll change it to TryGetMemoryContainerName

glahaye (approved):

dehoward (approved):

TaoChenOSU (approved):

PR #415, opened 2023-09-28
Minor updates before removing auth env variables

dehoward:
### Motivation and Context

this PR makes small changes which will be required in a follow-up PR that removes the AAD env variables from the frontend.

### Contribution Checklist

- [X] The code builds clean without any errors or warnings
- [X] The PR follows the [Contribution Guidelines](https://github.com/microsoft/chat-copilot/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/chat-copilot/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [X] All unit tests pass, and I have added new tests where possible
- [X] I didn't break anyone :smile:


alliscode (dismissed):

webapp/src/components/header/settings-dialog/SettingSection.tsx
```diff
@@ -56,7 +57,9 @@ export const SettingSection: React.FC<ISettingsSectionProps> = ({ setting, conte
                             <Switch
                                 label={feature.label}
                                 checked={feature.enabled}
-                                disabled={feature.inactive}
+                                disabled={
+                                    !!feature.inactive || (key === FeatureKeys.MultiUserChat && !AuthHelper.isAuthAAD())
```
glahaye, line 61: !!feature.inactive???  Isn't that the same as feature.inactive, or does it mean something special in JavaScript that I don't know about?

webapp/src/libs/services/BaseService.ts
```diff
@@ -12,21 +12,26 @@ interface ServiceRequest {
 
 const noResponseBodyStatusCodes = [202, 204];
 
+export const BackendServiceUrl = process.env.REACT_APP_BACKEND_URI ?? window.origin;
```
glahaye, line 15: Wondering whether we should check for empty strings in addition to null ones...

webapp/src/components/header/settings-dialog/SettingSection.tsx
dehoward: in this case it should be, it's ESLint that is confused here:
image

it is suggesting a nullish coalescing operator, I'm guessing because it sees that the variable is `boolean | undefined`, but that will incorrectly change the logic when `feature.inactive` is `false`, since the 2nd condition won't be evaluated.
 adding a `!!` forces the type to be just `boolean`, allowing ESLint to play nicely.

webapp/src/libs/services/BaseService.ts
dehoward: `checkEnv.ts` is currently handling this for us, but in my next PR that file will be going away. 

so now instead of checking for missing variables on start-up, we will first try to fetch the auth info with the URL that we have, and if that fails then we can point out that the `REACT_APP_BACKEND_URI` is empty if that is the case.

glahaye (dismissed):

glahaye (approved):

PR #416, opened 2023-09-28
Fix repository TryFindByIdAsync

glahaye:
### Motivation and Context
Repository TryFindByIdAsync would not try to find anything if no callback was passed to it, hence always returning true, even when no entries were found!

### Description
Always doing a true search because executing any callback with the result.

### Contribution Checklist

### DIFF
--- a/webapi/Storage/Repository.cs
+++ b/webapi/Storage/Repository.cs
@@ -29,7 +29,7 @@ public class Repository<T> : IRepository<T> where T : IStorageEntity
     {
         if (string.IsNullOrWhiteSpace(entity.Id))
         {
-            throw new ArgumentOutOfRangeException(nameof(entity.Id), "Entity Id cannot be null or empty.");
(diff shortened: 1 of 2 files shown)


webapi/Storage/Repository.cs
```diff
@@ -52,7 +52,10 @@ public async Task<bool> TryFindByIdAsync(string id, string? partition = null, Ac
     {
         try
         {
-            callback?.Invoke(await this.FindByIdAsync(id, partition ?? id));
+            T? found = await this.FindByIdAsync(id, partition ?? id);
+
+            callback?.Invoke(found);
```
dehoward, line 57: why does saving the call in a variable change the logic?

glahaye: Because if callback was null, Invoke() wouldn't be called and, thus, FindByAsync() neither

TaoChenOSU (approved):

PR #424, opened 2023-09-29
Plugin refactor

TaoChenOSU:
### Motivation and Context

Currently, plugin development using Chat Copilot has the following drawbacks:
1. CC requires the end user to configure the plugins and once configured, the plugins are transient, meaning that they disappear when the user refreshes the app. 
2. Plugins are scoped to the user who configured them, meaning that all users who want to use plugins must independently configure them.

### Description

This PR addresses the above issues by:
1. Move plugin configuration to the backend. Plugins are configured and added in the backend.
2. Save the states of plugins within chat sessions.
3. Create a new plugin call WebSearcher that performs Bing search. This plugin can be deployed to Azure.
4. Move the Klarna plugin to the backend.

This is not a breaking change.

This will be the first of a couple more PRs to come. The follow-up PRs will move the existing plugins that are still managed by the frontend to this new pattern.

image

### Contribution Checklist


dehoward:
@TaoChenOSU 61 files changed can be a lot to take in at once, is there any chance this PR can be broken up into 1 or more smaller PRs?

TaoChenOSU:
> @TaoChenOSU 61 files changed can be a lot to take in at once, is there any chance this PR can be broken up into 1 or more smaller PRs?

Yes, I will break this into multiple PRs.

plugins/web-searcher/Models/BingSearchResponse.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json.Serialization;
+
+namespace WebSearcher.Models;
```
alliscode, line 5: Nit: should this be Plugins.WebSearcher.Models?

webapi/Controllers/PluginController.cs
```diff
@@ -0,0 +1,110 @@
...
+        this._sessionRepository = sessionRepository;
+    }
+
+    /// <summary>
+    /// Fetches a plugin's manifest.
+    /// </summary>
+    /// <param name="manifestDomain">The domain of the manifest.</param>
+    /// <returns>The plugin's manifest JSON.</returns>
+    [HttpGet]
+    [Route("getPluginManifest")]
```
alliscode, line 49: We don't need "Get" in the route since it's a GET request. Also, pluginManifest should be plural since it's a collection whose Id is specified by manifestDomain. So the route should be `pluginManifests/{domain}` where domain is what is currently in the query.

webapi/Controllers/PluginController.cs
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        return this.Ok(await response.Content.ReadAsStringAsync());
+    }
+
+    /// <summary>
+    /// Enable or disable a plugin for a chat session.
+    /// </summary>
+    [HttpPut]
+    [ActionName("SetPluginStateAsync")]
```
alliscode, line 71: To make this route RESTfull, I think it should be modeled as a mutation of a resource, either the plugin resource or the chat resource. So either `PUT plugins/{pluginId}/chats/{chatId}` or `PUT chats/{chatId}/plugins/{pluginId}`

plugins/README.md
```diff
@@ -0,0 +1,49 @@
...
+- [WebSearcher](./web-searcher/README.md): A plugin that allows the chat bot to perform Bing search.
+- More to come. Stay tuned!
+
+## Third Party plugins
+
+You can also configure Chat Copilot to use third party plugins.
+
+> All no-auth plugins will be supported.
+
+> All service-level-auth plugins will be supported.
```
teresaqhoang, line 20:
```suggestion
> All service-level- and user-level- auth plugins will be supported.
```

plugins/README.md
```diff
@@ -0,0 +1,49 @@
...
+> OAuth plugins will NOT be supported.
+
+Read more about plugin authentication here: [Plugin authentication](https://platform.openai.com/docs/plugins/authentication)
+
+## Plugin Configuration in Chat Copilot
+
+### Prerequisites
+
+1. The name of your plugin. This should be identical to the `NameForHuman` in your plugin manifest.
+2. Url of your plugin.
```
teresaqhoang, line 31: Specify whether this should be the manifest url or the url pointing to the openapi spec. 

It would also be good to all out the requirements for plugin name and other fields defined by open ai here: https://platform.openai.com/docs/plugins/getting-started/plugin-manifest

plugins/shared/PluginApi.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// The API specification
+    /// </summary>
+    public string Type { get; set; } = "openapi";
+
+    /// <summary>
+    /// URL used to fetch the specification
+    /// </summary>
+    public string Url { get; set; } = string.Empty;
```
teresaqhoang, line 18: This can be done in a later PR but we should store the URL for the manifest, not the Open API spec.

I know we don't use any other fields from the manifest now, but there's work scheduled to leverage the other properties from the manifest (i.e., name for model and description for model) to make plugin discovery / use more effective.

plugins/shared/PluginAuth.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        /// <summary>
+        /// The API key
+        /// </summary>
+        public string OpenAI { get; set; } = string.Empty;
+    }
+
+    /// <summary>
+    /// The authentication schema
+    /// Supported values: none, service_http
```
teresaqhoang, line 25: why not user_http?

plugins/shared/PluginManifest.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// </summary>
+    [JsonPropertyName("name_for_human")]
+    public string NameForHuman { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Description of the plugin that the model will use
+    /// </summary>
+    [JsonPropertyName("description_for_model")]
+    public string DescriptionForModel { get; set; } = string.Empty;
+    /// <summary>
```
teresaqhoang, line 35: nit: missing line between properties here 
```suggestion

    /// <summary>
```

plugins/shared/PluginManifest.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json.Serialization;
+
+namespace PluginShared;
+
+/// <summary>
+/// This class represents the OpenAI plugin manifest.
```
teresaqhoang, line 8: Would be helpful to link the OpenAI manifest definition here: https://platform.openai.com/docs/plugins/getting-started/plugin-manifest

plugins/web-searcher/PluginEndpoint.cs
```diff
@@ -0,0 +1,195 @@
...
+        this._logger = logger;
+    }
+
+    /// <summary>
+    /// Gets the plugin manifest.
+    /// </summary>
+    /// <param name="req">The http request data.</param>
+    /// <returns>The manifest in Json</returns>
+    [Function("WellKnownAIPlugin")]
+    public async Task<HttpResponseData> WellKnownAIPlugin(
```
teresaqhoang, line 47: nit: Makes more clear to what we're requesting (the manifest rather than the plugin itself)
```suggestion
    public async Task<HttpResponseData> WellKnownAIPluginManifest(
```

webapi/Controllers/PluginController.cs
teresaqhoang: @alliscode this method isn't loading the plugin manifest from a stored collection. It's just making a fetch call to the URL domain that's passed it, essentially, loading the JSON from a webpage, so maybe something like `domains/{manifestDomain}/pluginManifest` would make more sense

webapi/Extensions/ServiceExtensions.cs
```diff
@@ -119,7 +170,7 @@ internal static IServiceCollection AddCorsPolicy(this IServiceCollection service
                     policy =>
                     {
                         policy.WithOrigins(allowedOrigins)
-                            .WithMethods("GET", "POST", "DELETE")
+                            .WithMethods("POST", "GET", "PUT", "DELETE")
```
teresaqhoang, line 173: Can you just use AllowAnyMethod() here?

webapi/README.md
```diff
@@ -73,6 +73,14 @@ Running the memory creation pipeline steps in different processes. This means th
 
 6. For more information and other options, please refer to the [memorypipeline](../memorypipeline/README.md).
 
+## Adding Plugins to the WebApi
+
+You can also add plugins that will be managed by the webapi (as opposed to manged by the webapp). Soon, all plugins will be managed by the webapi.
```
teresaqhoang, line 78: managed
```suggestion
You can also add plugins that will be managed by the webapi (as opposed to managed by the webapp). Soon, all plugins will be managed by the webapi.
```

teresaqhoang: This section should also be consolidated with the `### Adding Custom Plugins` section below

webapp/src/redux/features/conversations/ConversationsState.ts
```diff
@@ -34,3 +34,9 @@ export interface ConversationSystemDescriptionChange {
     id: string;
     newSystemDescription: string;
 }
+
+export interface PluginStateChangedPayload {
```
teresaqhoang, line 38: nit:
```suggestion
export interface UpdatePluginStatePayload {
```

webapp/src/redux/features/conversations/conversationsSlice.ts
```diff
@@ -184,6 +185,23 @@ export const conversationsSlice = createSlice({
             frontLoadChat(state, id);
             return;
         },
+        changePluginState: (state: ConversationsState, action: PayloadAction<PluginStateChangedPayload>) => {
```
teresaqhoang, line 188: nit: 
```suggestion
        updatePluginState: (state: ConversationsState, action: PayloadAction<UpdatePluginStatePayload>) => {
```

plugins/shared/PluginApi.cs
TaoChenOSU: This is the Url required by the manifest. We don't store this. this class is used to generate the manifest by code as opposed to a json file..

plugins/shared/PluginAuth.cs
TaoChenOSU: Added.

webapi/Extensions/ServiceExtensions.cs
TaoChenOSU: This will include changes that are not needed by this PR.

webapi/README.md
TaoChenOSU: I have made it more explicit that the new section is for OpenAI plugins. This should make it clearer that it's different than native & semantic functions/skills/plugins.

webapi/Options/PluginOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// The name of the plugin.
+    /// </summary>
+    public string Name { get; set; } = string.Empty;
+
+    /// <summary>
+    /// The url of the plugin.
+    /// </summary>
+    public Uri Url { get; set; } = new Uri("http://localhost");
```
teresaqhoang, line 20: To keep definitions aligned, this should be named ManifestDomain since we're following the standard defined here: https://platform.openai.com/docs/plugins/production/domain-verification-and-security

On the front end, we also refer to this URL as the manifest domain through the wizard 

```suggestion
    public Uri ManifestDomain { get; set; } = new Uri("http://localhost");
```

teresaqhoang (changes requested):

alliscode (dismissed):

webapp/src/components/open-api-plugins/PluginGallery.tsx
```diff
@@ -93,7 +89,12 @@ export const PluginGallery: React.FC = () => {
                         setHostedPlugins((hostedPlugins) => [...hostedPlugins, newHostedPlugin]);
                     })
                     .catch((error: Error) => {
-                        dispatch(addAlert({ message: error.message, type: AlertType.Error }));
+                        dispatch(
+                            addAlert({
+                                message: `Failed fetch hosted plugins: ${error.message}`,
```
glahaye, line 94: Failed TO fetch...

Also, this is done for each plugin, right?  So it should be Failed to fetch hosted plugin (no s at then) and the error message would be better if it include info as to _which_ plugin failed.

teresaqhoang (dismissed):

webapp/src/components/open-api-plugins/PluginGallery.tsx
TaoChenOSU: Good catch!

teresaqhoang (approved):

PR #427, opened 2023-09-30
Fetch AAD authentication info from backend

dehoward:
### Motivation and Context

this PR prepares the frontend for #377 and removes the need for any AAD configuration environment variables.

### Description
- removes `REACT_APP_AUTH_TYPE` and all variables starting with `REACT_APP_AAD_`
- calls the `/authConfig` endpoint when the app first loads and if needed, renders the `MsalProvider` using the fetched config.
- updates workflows and deployment scripts accordingly

### Contribution Checklist

- [X] The code builds clean without any errors or warnings
- [X] The PR follows the [Contribution Guidelines](https://github.com/microsoft/chat-copilot/blob/main/CONTRIBUTING.md) and the [pre-submission formatting script](https://github.com/microsoft/chat-copilot/blob/main/CONTRIBUTING.md#development-scripts) raises no violations
- [X] All unit tests pass, and I have added new tests where possible
- [X] I didn't break anyone :smile:


webapp/src/index.tsx
```diff
@@ -28,49 +27,54 @@ document.addEventListener('DOMContentLoaded', () => {
...
+                                </MsalProvider>
+                            </ReduxProvider>
+                        </React.StrictMode>,
+                    );
+                }
+            })
+            .catch((e: unknown) => {
+                if (e instanceof TypeError) {
+                    // fetch() will reject with a TypeError when a network error is encountered.
+                    store.dispatch(setAuthConfig(null));
```
glahaye, line 66: Why have two separate cases, one with null, the other with undefined?

webapp/.env.example
```diff
@@ -2,14 +2,6 @@
 # If you add any new required variables, make sure you update the variables list in the checkEnv.ts file as well.
```
glahaye, line 2: Remove this comment as it doesn't apply anymore,

webapp/src/index.tsx
glahaye: Oh, I just saw the code in App.tsx...  Maybe put a comment here too?

webapp/src/App.tsx
```diff
@@ -205,7 +221,14 @@ const Chat = ({
             {appState === AppState.ErrorLoadingUserInfo && (
                 <Error text={'Oops, something went wrong. Please try signing out and signing back in.'} />
             )}
-            {appState === AppState.LoadingChats && <Loading text="Loading Chats..." />}
+            {appState === AppState.ErrorLoadingAuthInfo && (
+                <Error text={'Oops, unable to authentication info. Please try refreshing the page.'} />
```
glahaye, line 225: Maybe "Unable to load authentication info[...]" or "Unable to authenticate user[...]"?

Also, I would remove the "Oops" from all the strings here.

glahaye (approved):

webapp/src/App.tsx
dehoward: ah yes, typo, it should be "Unable to **load** ...".

glahaye (approved):

dehoward:
needs to be tested in Azure before merging

alliscode (approved):

webapp/src/App.tsx
```diff
@@ -50,73 +50,95 @@ export const useClasses = makeStyles({
...
     const isAuthenticated = useIsAuthenticated();
 
     const chat = useChat();
     const file = useFile();
 
+    useEffect(() => {
+        // if the auth info is being loaded, change the state if:
+        //      1. the `authConfig` is falsy, meaning an error occurred during load, or
+        //      2. the `authConfig` has keys, meaning it was loaded successfully.
+        if (appState === AppState.LoadingAuthInfo && (!authConfig || Object.keys(authConfig).length)) {
```
teresaqhoang, line 79: This setup bypasses the ProbeForBackend state altogether unless we hit a specific error case, but we need to hit the ProbeForBackend state at least once so the maintenceCheck can be hit. 

It might make more sense to Load the AuthInfo in the BackendProbe component in this series of events:
1. LoadAuthInfo
2. then Maintenace Check

If both 1 and 2 pass, then you can bypass the /healthz check there (which is how it's setup now where MaintainenceCheck is queried and the /healthz probe is bypassed if /maintencestatus is hit)

teresaqhoang (changes requested):

glahaye:
> needs to be tested in Azure before merging

It worked fine when I tested it on Azure!

webapp/src/App.tsx
dehoward: thanks for calling out the maintenance check. I've updated `App` to always begin on the `ProbeForBackend` state, can you take another look?

I ran into issues loading the auth info in `BackendProbe`, since the auth info is used to render the `MsalProvider`, which needs to sit above the `App` in the React tree for the hooks to work. I experimented with having the `BackendProbe` be the top-level- component in `index.tsx`, but it led to too much polling and some other weirdness.

webapi/Controllers/MaintenanceController.cs
```diff
@@ -53,7 +53,7 @@ public class MaintenanceController : ControllerBase
 
         var migrationStatus = await migrationMonitor.GetCurrentStatusAsync(cancellationToken);
 
-        if (migrationStatus != ChatMigrationStatus.None)
+        if (migrationStatus == ChatMigrationStatus.Upgrading)
```
glahaye, line 56: Why change this? What about the RequiresUpgrade state?

glahaye: Moving this to other convo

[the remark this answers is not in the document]
glahaye: Why change this? What about the RequiresUpgrade state?

Also, maybe we could simplify things and make that flag part of ServiceInfo instead of having to make a call for ServiceInfo and another for Maintenance status.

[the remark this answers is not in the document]
dehoward: because of this comment on the frontend:
image

there is an existing assumption that when an object is returned from the `/maintenanceStatus` API, that the backend is in maintenance mode, but this is not the case when the status is `RequiresUpgrade` (which mine was in and is why I noticed this).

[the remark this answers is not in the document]
crickman: I think "RequiresUpgrade" is also valid maintenance status.  Not sure how this change affects maintenance data-flow.

webapp/src/components/views/BackendProbe.tsx
```diff
@@ -18,55 +20,61 @@ interface IMaintenance {
...
-                result
-                    .json()
+            const fetchMaintenanceAsync = () =>
+                fetch(migrationUrl)
+                    .then((response) => response.json())
                     .then((data) => {
-                        // Body has payload.  This means the app is in maintenance
+                        // Body has payload. This means the app is in maintenance
                         setModel(data as IMaintenance);
+                        return true;
```
teresaqhoang, line 58: Should this be returning true here? If the body returns a payload, that's enough to verify that the backend is up so we should not need to probe.

webapi/Controllers/MaintenanceController.cs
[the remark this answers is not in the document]
teresaqhoang: Doesn't the `RequiresUpgrade` state trigger the upgrade automatically?

webapi/Controllers/MaintenanceController.cs
teresaqhoang, line 1: Can you remove the authInfo dependency injection here? It might get confusing that this service has a dependency on auth when it doesn't, especially now that this call can be made before auth is loaded.

webapp/src/libs/auth/AuthHelper.ts
```diff
@@ -9,36 +9,27 @@ import {
...
-// It must match what is defined in PassthroughAuthenticationHandler.cs on the backend.
-export const DefaultChatUser: IChatUser = {
-    id: 'c05c61eb-65e4-4223-915a-fe72b0c9ece1',
-    emailAddress: 'user@contoso.com',
-    fullName: 'Default User',
-    online: true,
-    isTyping: false,
-};
-
-export const DefaultActiveUserInfo: ActiveUserInfo = {
```
teresaqhoang, line 32: Did you mean to remove this? Isn't it still referenced as the default in AuthHelper?

webapi/Controllers/MaintenanceController.cs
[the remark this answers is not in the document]
dehoward: I'm going to leave it as-is for now to not block the PR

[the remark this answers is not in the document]
teresaqhoang: We should either revert this change or confirm that other ChatMigrationStatus states, like `RequiresUpgrade` don't have an effect, otherwise, this will cause a regression.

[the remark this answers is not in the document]
dehoward: sorry that's what I meant, that I will revert the change.. on the frontend, it doesn't currently make a big difference since the `isMaintenance` flag isn't set

teresaqhoang (approved):

PR #444, opened 2023-10-02
Update deploy-to-azure button description

TaoChenOSU:
### Motivation and Context

The deploy-to-azure button says it will deploy the latest CC binaries, which is not correct and has created a lot of confusion.

### Description
Update the description of the button and we should make updating this description as part of out release workflow.

### Contribution Checklist

### DIFF
--- a/scripts/deploy/README.md
+++ b/scripts/deploy/README.md
@@ -124,7 +124,7 @@ You can also deploy the infrastructure directly from the Azure Portal by clickin
 
 [![Deploy to Azure](https://aka.ms/deploytoazurebutton)](https://aka.ms/sk-deploy-existing-azureopenai-portal)
 
-> This will automatically deploy the most recent release of Chat Copilot binaries ([link](https://github.com/microsoft/chat-copilot/releases)).
+> This will automatically deploy the [0.5 release](https://github.com/microsoft/chat-copilot/releases/tag/0.5) of the CopilotChat infrastructures and backend binaries.
 
 > To find the deployment name when using `Deploy to Azure`, look for a deployment in your resource group that starts with `Microsoft.Template`.
 


scripts/deploy/README.md
```diff
@@ -125,7 +125,7 @@ You can also deploy the infrastructure directly from the Azure Portal by clickin
 
 [![Deploy to Azure](https://aka.ms/deploytoazurebutton)](https://aka.ms/sk-deploy-existing-azureopenai-portal)
 
-> This will automatically deploy the most recent release of CopilotChat backend binaries ([link](https://github.com/microsoft/chat-copilot/releases)).
+> This will automatically deploy the [0.4 release](https://github.com/microsoft/chat-copilot/releases/tag/0.4) of the CopilotChat infrastructures and backend binaries.
```
glahaye, line 128: Writing this way means we won't have to change this every build!

```suggestion
> This will deploy the latest [release](https://github.com/microsoft/chat-copilot/tags) of the Chat Copilot to Azure.
```

TaoChenOSU: We are not always deploying the latest though.

glahaye: Argh...   You are right...  But I would hate to have to maintain the correspondance between the version here and the version online.  Because then we'll have to keep changing this file and/or it will drift...

Maybe "This will deploy the latest stable release (not necessarily the latest code from the main branch) of the Chat Copilot to Azure."???

TaoChenOSU: Well, the reality is we have 0.5 but this button is still deploying 0.4. An alternative would be to mark a release as the stable release. We will still have to make sure the version here matches the stable version regardless.

dehoward: related question: why are we not always deploying the latest release? if it's not stable, why release it? it seems like we should we waiting until the build is stable before doing an official release.

glahaye: That's a VERY good question,  I was just talking about that with Tao yesterday. I'll have a talk with Ben regarding this.

Right now, we cut new builds right after breaking changes are introduced (i.e, at a moment where the build is more likely to be unstable) whereas we should probably release builds right BEFORE putting in breaking changes...

PR #460, opened 2023-10-05
Fix hanging when status is RequiresUpgrade

dehoward:
### Motivation and Context

when the `ChatMigrationStatus` is set to `RequiresUpgrade`, the frontend will hang while seemingly nothing happens in the backend. changing the backend to only send the `MaintenanceResult` when an upgrade is happening fixes this. 

### Contribution Checklist

### DIFF
--- a/webapi/Controllers/MaintenanceController.cs
+++ b/webapi/Controllers/MaintenanceController.cs
@@ -49,7 +49,7 @@ public class MaintenanceController : ControllerBase
 
         var migrationStatus = await migrationMonitor.GetCurrentStatusAsync(cancellationToken);
 
-        if (migrationStatus != ChatMigrationStatus.None)
+        if (migrationStatus == ChatMigrationStatus.Upgrading)
         {
             result =
                 new MaintenanceResult
--- a/webapp/src/App.tsx
+++ b/webapp/src/App.tsx
@@ -166,6 +166,15 @@ const Chat = ({
     appState: AppState;
     setAppState: (state: AppState) => void;
 }) => {
+    const onBackendFound = React.useCallback(() => {
+        setAppState(
+            AuthHelper.isAuthAAD()
+                ? // if AAD is enabled, we need to set the active account before loading chats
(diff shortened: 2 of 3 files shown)


webapi/Controllers/MaintenanceController.cs
```diff
@@ -49,7 +49,7 @@ public class MaintenanceController : ControllerBase
 
         var migrationStatus = await migrationMonitor.GetCurrentStatusAsync(cancellationToken);
 
-        if (migrationStatus != ChatMigrationStatus.None)
+        if (migrationStatus == ChatMigrationStatus.Upgrading)
```
glahaye, line 52: I thought we WANTED to hang and display an error message when an upgrade is required???

dehoward: let's figure this out. imo we shouldn't block the entire chat experience if it works, and should probably show this error as a status alert..

glahaye: @crickman is our man, here. He's the one who put this in so he would know how we should react to it.

glahaye: Oh, and again, maybe we should make this a field in the ServiceInfo response to minimize the number of calls we make to ther server??

crickman (approved):

webapi/Controllers/MaintenanceController.cs
dehoward: @glahaye just had a sync with @crickman, we discovered that in local mode the backend will _always_ be in the `RequiresUpgrade` state since the ACS index will not exist. he is going to look into this further. 
we agreed that we should only show the "Migrating memory..." message when the migration is actually occurring, since you may be in a state where you need it but it isn't currently happening, in which case you should still be able to chat.

glahaye: If he's happy, I'm happy :)

glahaye (approved):

PR #476, opened 2023-10-06
Volatile memory singleton for memory "migration"

crickman:
### Motivation and Context

Muliple instances of VolatileMemory do not share state.

### Contribution Checklist

### DIFF
--- a/webapi/Services/SemanticKernelProvider.cs
+++ b/webapi/Services/SemanticKernelProvider.cs
@@ -23,7 +23,7 @@ namespace CopilotChat.WebApi.Services;
 /// </summary>
 public sealed class SemanticKernelProvider
 {
-    private static IMemoryStore? _memoryStore;
+    private static IMemoryStore? _volatileMemoryStore;
 
     private readonly IServiceProvider _serviceProvider;
     private readonly IConfiguration _configuration;
@@ -163,8 +163,8 @@ public sealed class SemanticKernelProvider
             {
                 case string x when x.Equals("SimpleVectorDb", StringComparison.OrdinalIgnoreCase):
                     // Maintain single instance of volitile memory.
-                    Interlocked.CompareExchange(ref _memoryStore, new VolatileMemoryStore(), null);
-                    return _memoryStore;
+                    Interlocked.CompareExchange(ref _volatileMemoryStore, new VolatileMemoryStore(), null);
+                    return _volatileMemoryStore;
 
                 case string x when x.Equals("Qdrant", StringComparison.OrdinalIgnoreCase):
                     var qdrantConfig = memoryOptions.GetServiceConfig<QdrantConfig>(this._configuration, "Qdrant");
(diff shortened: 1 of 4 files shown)


webapi/Services/SemanticKernelProvider.cs
```diff
@@ -159,7 +162,9 @@ IMemoryStore CreateMemoryStore()
             switch (memoryOptions.Retrieval.VectorDbType)
             {
                 case string x when x.Equals("SimpleVectorDb", StringComparison.OrdinalIgnoreCase):
-                    return new VolatileMemoryStore();
+                    // Maintain single instance of volitile memory.
```
glahaye, line 165: Typo in volatile

```suggestion
                    // Maintain single instance of volatile memory.
```

webapi/Services/SemanticKernelProvider.cs
```diff
@@ -22,6 +23,8 @@ namespace CopilotChat.WebApi.Services;
 /// </summary>
 public sealed class SemanticKernelProvider
 {
+    private static IMemoryStore? _memoryStore;
```
glahaye, line 26: Maybe rename this guy _volatileMemoryStore since it only applies to volatile memory...

webapi/Services/SemanticKernelProvider.cs
```diff
@@ -159,7 +162,9 @@ IMemoryStore CreateMemoryStore()
             switch (memoryOptions.Retrieval.VectorDbType)
             {
                 case string x when x.Equals("SimpleVectorDb", StringComparison.OrdinalIgnoreCase):
-                    return new VolatileMemoryStore();
+                    // Maintain single instance of volitile memory.
+                    Interlocked.CompareExchange(ref _memoryStore, new VolatileMemoryStore(), null);
```
glahaye, line 166: I notice you are a big fan of CompareExchange.  That's all good.  But for creating singletons, I like using Lazy<t>. If you don't know about Lazy, check it out. I find it cleaner to create singletons.

This is just a public information announcement,  Feel free to ignore,

webapi/Services/SemanticKernelProvider.cs
crickman: Of course...always.  One day I'll learn 👍

webapi/Services/SemanticKernelProvider.cs
crickman: Lazy had some well known problems...not sure if those got addressed in .net core, but there's a historical reason people avoid Lazy.

crickman: Yeah, just did double-check...not a fan of `Lazy`.  I think the highest bar here would be to define VolatileMemoryStore as a singleton and then sp.GetRequiredService()...but because of the DI layering, its a bit of a pita here (and just not worth it).

FWIW - This code will be removed after a few more releases.

glahaye (approved):

Issue #478, opened 2023-10-08
MySQL implementation of DbContext/IStorageContext


Ryangr0:
I would like for there to be an option in the persistency layer that is not Cosmos (because it's a paid service) and not Filesystem or Volatile for obvious reasons. I got this to work in my local project, with some caveats (repositories, and therefore seemingly IStorageContext, is registered to be a singleton. That does not play well with the transactional nature of MySQL). I'm making this issue to check to see if there is an audience for this functionality.

Also a question. What has been decided about EF for the data model layer?

Rainson12:
Hi @Ryangr0,
I have already implemented a EF Layer for mariadb which should also work for mysql. Have a look at https://github.com/microsoft/chat-copilot/issues/242#issuecomment-1740899503

TaoChenOSU:
Hello @Ryangr0, 

Thanks for your post!

You are more than welcome to submit PRs if you think it's ready!

matthewbolanos:
Closing this issue since the Chat Copilot team is not ready to merge a feature like this in.

Issue #242, opened 2023-08-23
Migrate CosmosDB code to Entity Framework


dluc:
I believe we could reduce the cost and complexity by using Entity Framework. That should allow to remove some code, not having to worry about security fixes and so on, plus EF is very mature and well known to most C#/DB devs, so we won't ask to learn a new/bespoke approach.

Some resources:
* https://learn.microsoft.com/ef/core/providers/cosmos/
* https://learn.microsoft.com/ef/core/providers/cosmos/limitations

Rainson12:
you might want to check https://github.com/Rainson12/chat-copilot/commit/7207c3e41791c19a282b3a4fe988a28e6ca477f4
I added EFcore support and use mariadb as storage. There might be 2 issues which i suboptimally fixed. 

1. the chatmessages has a property of type Dictionary (https://github.com/microsoft/chat-copilot/blob/3c8dab396b442f02b307dcb2aa39fc6d2c64b01e/webapi/Models/Storage/ChatMessage.cs#L117) which can not be mapped to an entity as there is no primary key and so entityframework doesnt like it. Also I added an empty constructor.
2. The setup of the DB could be optimized. Due to the way on how the storageContexts are set, there will be 4 instances created (see https://github.com/Rainson12/chat-copilot/commit/7207c3e41791c19a282b3a4fe988a28e6ca477f4#diff-373907b269e2e6425302f31de796a82cde423beb713f62a7a73ed6ffefa08eacR204). Each instance only holds a dbset of the required entity, this allows for adding new entities without having to adjust the dbcontext however it has the downside of not being able to setup the database using EnsureCreated() as only the first table would be created, for every subsequent instantiation it wouldnt work, as the db is already created however the tables are not. I therefore check for the existance of the table and if it doesnt exist, create it using rawSQL generated by entityframework.

Ryangr0:
Can we come together on this? I'd love to see your PR @Rainson12 . I basically did the following for mysql

```
case ChatStoreOptions.ChatStoreType.MySql:
                {
                    if (chatStoreConfig.MySql == null)
                    {
                        throw new InvalidOperationException("ChatStore:MySql is required when ChatStore:Type is 'MySql'");
                    }

                    services.AddDbContext<MySqlDbContext>(
                        options =>
                            options.UseMySql(
                                chatStoreConfig.MySql.ConnectionString,
                                new MariaDbServerVersion(new Version(11, 1, 1))
[...]

PR #500, opened 2023-10-14
Introduce Entity Framework & implement MySQL IContextStorage

Ryangr0:
### Motivation and Context

Original issue: #478 

Some background:
I want/need to persist the chat in something other than Cosmos or the filesystem. I'm aware that C# has a long history of ORM usage, so I thought it would make sense to implement MySQL for this, and introduce EF to make sure the datamodel layer is easy to change.

@Rainson12 has also made an effort in the same vein, for MariaDB, as seen here #242

So now I know of 3 people that want this, and I haven't found anyone so far that's opposed to the idea of EF.

### Description

This has serious consequences, during local testing I ran into a lot of problems surrounding concurrency. That meant that I had to change classes that aren't directly related to the change I'm trying to make. 

Also, right now there seems to be some kind of hack surrounding the scope for the Cosmos implementation that I'd like to get rid of as well.

I'd like some help with this! So if you want to help out, let me know. I don't really know all that much about EF itself, and I want to make sure we get this right the first time.

### Contribution Checklist

### DIFF
--- a/webapi/Extensions/ServiceExtensions.cs
+++ b/webapi/Extensions/ServiceExtensions.cs
@@ -239,6 +239,35 @@ public static class CopilotChatServiceExtensions
                 break;
             }
 
+            case ChatStoreOptions.ChatStoreType.MySql:
+            {
+                if (chatStoreConfig.MySql == null)
+                {
+                    throw new InvalidOperationException("ChatStore:MySql is required when ChatStore:Type is 'MySql'");
+                }
+
+                services.AddDbContext<MySqlDbContext>(
+                    options =>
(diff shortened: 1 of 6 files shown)


Ryangr0:
@microsoft-github-policy-service agree company="WebGrip"

Ryangr0:
That's too bad. I put some effort into this, there are multiple people that want this and I was told to make a PR on the issue I created. It's not about showing off, it's about supporting normal industry standards for persistency. Forcing cosmos is not a good look, just sayin'. Thinking about the data model in terms of ORM would be helpful as well, I think. But it's your decision :)

PR #524, opened 2023-10-19
Make deploy-webapi do all it needs to by default

glahaye:
### Motivation and Context
Some switches need to be included at the command line to ensure proper modification of the AAD app registration and addition of our URIs to the plugin CORS. These steps should be done by default.

### Description
Make all required steps run by default.
Fixed population of frontend version info

### Contribution Checklist

### DIFF
--- a/.github/workflows/copilot-deploy-backend.yml
+++ b/.github/workflows/copilot-deploy-backend.yml
@@ -81,4 +81,4 @@ jobs:
 
       - name: "Deploy"
         run: |
-          scripts/deploy/deploy-webapi.sh -p "${{ github.workspace }}/${{inputs.ARTIFACT_NAME}}/webapi.zip" -d ${{inputs.DEPLOYMENT_NAME}} -s ${{secrets.AZURE_SUBSCRIPTION_ID}} -rg ${{vars.CC_DEPLOYMENT_GROUP_NAME}} --register-cors
+          scripts/deploy/deploy-webapi.sh -p "${{ github.workspace }}/${{inputs.ARTIFACT_NAME}}/webapi.zip" -d ${{inputs.DEPLOYMENT_NAME}} -s ${{secrets.AZURE_SUBSCRIPTION_ID}} -rg ${{vars.CC_DEPLOYMENT_GROUP_NAME}} --skip-app-registration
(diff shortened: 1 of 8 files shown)


scripts/deploy/deploy-webapi.ps1
```diff
@@ -27,12 +27,12 @@ param(
     $PackageFilePath = "$PSScriptRoot/out/webapi.zip",
 
     [switch]
-    # Switch to add our URIs in app registration's redirect URIs if missing
-    $EnsureUriInAppRegistration,
+    # Don't attempt to add our URIs in app registration's redirect URIs
```
glahaye, line 30: Yes, this changes the existing logic.  But given this the current logic has only worked from this week and that people are already complaining about how it is incomplete, I'm fixing it now to kill this problem right away.

TaoChenOSU: nit: maybe worth mentioning it's the WebApp's or the frontend's end registration.

.github/workflows/copilot-deploy-backend.yml
```diff
@@ -81,4 +81,4 @@ jobs:
 
       - name: "Deploy"
         run: |
-          scripts/deploy/deploy-webapi.sh -p "${{ github.workspace }}/${{inputs.ARTIFACT_NAME}}/webapi.zip" -d ${{inputs.DEPLOYMENT_NAME}} -s ${{secrets.AZURE_SUBSCRIPTION_ID}} -rg ${{vars.CC_DEPLOYMENT_GROUP_NAME}} --register-cors
+          scripts/deploy/deploy-webapi.sh -p "${{ github.workspace }}/${{inputs.ARTIFACT_NAME}}/webapi.zip" -d ${{inputs.DEPLOYMENT_NAME}} -s ${{secrets.AZURE_SUBSCRIPTION_ID}} -rg ${{vars.CC_DEPLOYMENT_GROUP_NAME}} --skip-app-registration
```
TaoChenOSU, line 84: If we do this, will the app get redirected back correctly after authentication?

scripts/deploy/deploy-webapi.ps1
glahaye: Done

TaoChenOSU (approved):

scripts/deploy/deploy-webapi.ps1
glahaye: Done

.github/workflows/copilot-deploy-backend.yml
glahaye: No change from previous behavior. This actually makes the absence of registration more obvious

PR #546, opened 2023-10-26
Update to "final" semantic-memory package

crickman:
### Motivation and Context
1. Semantic-Memory is being renamed and repackage.
1. Semantic-Kernel is also going through march to release (currently beta3)
1. CC and SM both have dependency on SK. 
1. Ideally CC and SM should update to the same SK beta (SM first and then CC)
1. This is the final available update of SM for CC prior to any SK beta update and includes functional and stabilization updates

### Contribution Checklist

### DIFF
--- a/memorypipeline/appsettings.json
+++ b/memorypipeline/appsettings.json
@@ -57,23 +57,23 @@
       // - Directory is the location where files are stored.
       //
       "SimpleFileStorage": {
-        "Directory": "../webapi/tmp-cache"
+        "Directory": "../tmp/cache"
       },
       //
       // File based queue for local/development use.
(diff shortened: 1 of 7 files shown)


memorypipeline/appsettings.json
```diff
@@ -57,23 +57,23 @@
       // - Directory is the location where files are stored.
       //
       "SimpleFileStorage": {
-        "Directory": "../webapi/tmp-cache"
+        "Directory": "../tmp/cache"
```
TaoChenOSU, line 60: We should also update the gitignore file

crickman: Great call...doing now!

PR #581, opened 2023-11-08
Update Chat-Copilot to Semantic Kernel  Beta4 and Kernel-Memory

crickman:
### Motivation and Context

Updating chat-copilot to semantic-kernel 1.0.0-beta4 and also updating semantic-memory reference to kernel-memory (which also references SK 1.0.0-beta4)

### Contribution Checklist


webapi/Controllers/ChatController.cs
```diff
@@ -179,8 +176,7 @@ public class ChatController : ControllerBase, IDisposable
 
         // Verify that the chat exists and that the user has access to it.
         ChatSession? chat = null;
-#pragma warning disable CA1508 // Avoid dead conditional code. It's giving out false positives on chat == null.
-        if (!(await chatSessionRepository.TryFindByIdAsync(chatId, callback: c => chat = c)) || chat == null)
```
glahaye, line 183: The old code seemed to imply we could get a ChatSession that would be null. Is that possible?  If so, we are no longer covering that case...

webapi/Controllers/ChatController.cs
```diff
@@ -236,8 +232,8 @@ public class ChatController : ControllerBase, IDisposable
 
         AskResult chatSkillAskResult = new()
         {
-            Value = result.Result,
-            Variables = result.Variables.Select(v => new KeyValuePair<string, string>(v.Key, v.Value))
+            Value = result.GetValue<string>() ?? string.Empty,
+            Variables = contextVariables.Select(v => new KeyValuePair<string, string>(v.Key, v.Value))
```
glahaye, line 236: Is that field even relevant now since it looks like we are just returning the same variables we were given?

webapi/Controllers/ChatController.cs
```diff
@@ -296,13 +292,13 @@ private async Task RegisterPlannerSkillsAsync(CopilotChatPlanner planner, Dictio
         if (openApiSkillsAuthHeaders.TryGetValue("JIRA", out string? JiraAuthHeader))
         {
             this._logger.LogInformation("Registering Jira plugin");
-            var authenticationProvider = new BasicAuthenticationProvider(() => { return Task.FromResult(JiraAuthHeader); });
+            var authenticationProvider = new BearerAuthenticationProvider(() => { return Task.FromResult(JiraAuthHeader); });
```
glahaye, line 295: Where will we get this token?

webapi/Services/SemanticKernelProvider.cs
```diff
@@ -179,10 +188,10 @@ IMemoryStore CreateMemoryStore()
                     return _volatileMemoryStore;
 
                 case string x when x.Equals("Qdrant", StringComparison.OrdinalIgnoreCase):
-                    var qdrantConfig = memoryOptions.GetServiceConfig<QdrantConfig>(this._configuration, "Qdrant");
+                    var qdrantConfig = memoryOptions.GetServiceConfig<QdrantConfig>(configuration, "Qdrant");
 
 #pragma warning disable CA2000 // Ownership passed to QdrantMemoryStore
-                    HttpClient httpClient = new(new HttpClientHandler { CheckCertificateRevocationList = true });
+                    HttpClient httpClient = new(new HttpClientHandler { CheckCertificateRevocationList = true }); // $$$
```
glahaye, line 194: What's the // $$$ about?

crickman: Comment to myself...will clean

webapi/Controllers/ChatController.cs
crickman: Yeah, I saw that too.  The nullability in the definition is due to how the TryFindByIdAsync is doing the assignment in a callback.  Let's just call it an antipattern that wasn't introduced here (out of scope).

webapi/Controllers/ChatController.cs
crickman: I'll be doing optimizaiton analysis subsequent to this change.  Right now the goal is to maintain status/quo functionality.

webapi/Controllers/ChatController.cs
crickman: Updated

glahaye (approved):

PR #587, opened 2023-11-08
Replace "Skills" term, as appropriate.

crickman:
### Motivation and Context

Since late May 2023, we've been transitioning away from the term `Skills `in favor of `Plugins` in the semantic-kernel ecosystem.  This completes that transition for chat-copiilot.

> NOTE: There is one definition where the `skill` term cannot be replaced since its linked to the Plan class definition in semantic-kernel.

image
image

### Contribution Checklist


webapi/Extensions/SemanticKernelExtensions.cs
```diff
@@ -33,7 +33,7 @@ internal static class SemanticKernelExtensions
     /// <summary>
     /// Delegate to register plugins with a Semantic Kernel
```
momuno, line 34: This says "register plugins", while line 36 change says "registerFunctions..."  Can this please be clarified?  Seeing the terms plugin and function used interchangeably elsewhere too.

crickman: Plugin has one or more functions.  The plug in registers functions that are called as planner steps.  Just trying to align with how the term is being used in semantic-kernel.

gitri-ms (approved):

glahaye (approved):

PR #588, opened 2023-11-09
Unblock Global Document Import

teresaqhoang:
### Motivation and Context

This PR adds support to upload a document to global CC scope from the client:
image

### Description

Mollie's changes in the Document Import Controller 
- Updated the return of this.OK() on global document import to return a ChatMessage object instead of a string.
- Refactored internal functions to private to clarify what parameters were being changed but what functions. (Previously all internal in scope.)

### Contribution Checklist


crickman (approved):

glahaye (approved):

jzingh98:
Would like to bump this

teresaqhoang:
Needs more testing ---

Currently seeing this error when uploading to global but wanted to get them started.
image

It seems as if the functionality works and the document is uploaded to the global scope successfully and this error is falsy. Need further debugging before this PR can merge.

momuno:
Global upload:
603910e5-9c0d-4993-84ee-a4a7d2cd4545
Local Upload:
a96bbec6-faee-4302-b72f-7264e652dc82

It appears the global upload response header returns plain text, but the local upload response header returns application json. I think this might be the issue, hence "... is not valid JSON"

momuno:
Changes updated the return of this.OK() to use a chatmessage object instead of a string.  
Also refactored internal functions to private to clarify what parameters were being changed but what functions.  (Previously all internal in scope.)

webapi/Controllers/DocumentController.cs
```diff
@@ -136,20 +142,22 @@ public class DocumentController : ControllerBase
...
-            var chatMessage = await this.TryCreateDocumentUploadMessage(chatId, documentMessageContent);
-
             // If chat message isn't created, it is still broadcast and visible in the documents tab.
             // The chat message won't, however, be displayed when the chat is freshly rendered.
 
             var userId = this._authInfo.UserId;
             await messageRelayHubContext.Clients.Group(chatId.ToString())
                 .SendAsync(ReceiveMessageClientCall, chatId, userId, chatMessage);
 
+            this._logger.LogInformation("Local upload chat message: {0}", chatMessage!.ToString());
```
glahaye, line 158: Only ever user the '!' operator when you are 100% certain it can't be null - In this case it actually can!

We probably want to return above (likely a 400) if chatMessage is null.

Then the '!' won't be needed.  Same with the other log.

crickman (dismissed):

webapi/Controllers/DocumentController.cs
```diff
@@ -136,20 +141,28 @@ public class DocumentController : ControllerBase
...
         DocumentMessageContent documentMessageContent = new();
 
-        var importResults = await ImportDocumentsAsync();
+        var importResults = await this.ImportDocumentsAsync(memoryClient, chatId, documentImportForm, documentMessageContent);
+
+        var chatMessage = await this.TryCreateDocumentUploadMessage(chatId, documentMessageContent);
+
+        if (chatMessage == null)
+        {
+            this._logger.LogInformation("chatMessage is null. Returning 400.");
```
glahaye, line 150:
```suggestion
            this._logger.LogWarning("Failed to create document upload message - {content}", documentMessageContent.ToString());
```

glahaye (approved):

PR #642, opened 2023-11-21
Fixing conversation filtering

teresaqhoang:
### Motivation and Context

This PR removes part of the isFiltering conversations view state, including an unnecessary filteredConversations field, since @TaoChenOSU's filters and sorts the conversations into the expected Latest and Older conversations lists.

Also, fixes a bug so the filter is applied to the client view name, rather than the actual saved name on the object since this is what our end user sees and would expect the search text to check against.

### Contribution Checklist

### DIFF
--- a/webapp/src/components/chat/chat-list/ChatList.tsx
+++ b/webapp/src/components/chat/chat-list/ChatList.tsx
@@ -83,7 +83,6 @@ const useClasses = makeStyles({
 });
 
 interface ConversationsView {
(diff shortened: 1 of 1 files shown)


teresaqhoang:
@threegitty350 take a look at this and see if it helps with the issue you're seeing

momuno:
I tested this locally and filtering works now.

momuno (approved):

PR #700, opened 2023-12-07
Use chatId from URL rather than from payload for chats

glahaye:
### Motivation and Context
The verify access to a chat, we use HandleRequest() with the chatId provided. Currently, we get this from the payload, which can differ from the chatId from the URL, which opens us to a security problem where a user could inject an arbitrary chatId in the payload, which doesn't match what's in the URL.

### Description
- Use chatId from URL and only from URL
- Add integrations test to validate this

### Contribution Checklist


integration-tests/ChatTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using CopilotChat.WebApi.Models.Response;
+using Xunit;
+using static CopilotChat.WebApi.Models.Storage.CopilotChatMessage;
+
+namespace ChatCopilotIntegrationTests;
+
+public class ChatTests : ChatCopilotIntegrationTest
+{
+    [Fact]
+    public async void ChatWithBot()
```
alliscode, line 17: Can you make this name more targeted? Maybe something like `ChatMessagePostSucceedsWithValidInput`.

alliscode (approved):

webapi/Controllers/ChatController.cs
```diff
@@ -415,6 +412,28 @@ private async Task RegisterPlannerHostedFunctionsUsedAsync(CopilotChatPlanner pl
         return;
     }
 
+    private static ContextVariables GetContextVariables(Ask ask, IAuthInfo authInfo, string chatId)
```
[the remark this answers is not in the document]
glahaye: It was to simplify, yes.

webapi/Controllers/ChatController.cs
```diff
@@ -415,6 +412,28 @@ private async Task RegisterPlannerHostedFunctionsUsedAsync(CopilotChatPlanner pl
...
+    {
+        const string UserIdKey = "userId";
+        const string UserNameKey = "userName";
+        const string ChatIdKey = "chatId";
+
+        var contextVariables = new ContextVariables(ask.Input);
+        foreach (var variable in ask.Variables)
+        {
+            if (variable.Key != UserIdKey && variable.Key != UserNameKey)
```
TaoChenOSU, line 424: I know this is copied from somewhere else. But what is the point of this condition? We are going to override the values anyways.

glahaye: Yes, this check is redundant.  I'll remove it.

TaoChenOSU (approved):

PR #817, opened 2024-02-23
Upgrade to Semantic Kernel 1.4

glahaye:
### Motivation and Context
Chat Copilot hasn't been upgraded to SK 1 and this hampers development.

### Description
Upgrade to SK 1.4

### Contribution Checklist


ezar:
Thanks!!!

glahaye:
> Do you remove planner's kernel?

Yes.  The idea is that OpenAI's function-calling can figure out when to call functions without requiring a plan to do so.

Let me know if you think the lack of explicit planner will be a problem.

ezar:
@glahaye I have tried to use the Karna and Websercher plugins, but since the plan does not activate, they do not launch these and OpenAI responds directly. Is this the expected workflow? 

I had to correct some exceptions that were thrown when trying to use the chat. For example, the duplication of the chatid in the list.

Thank you very much because thanks to this project I can understand much more about the functioning of SK.

glahaye:
> @glahaye I have tried to use the Karna and Websercher plugins, but since the plan does not activate, they do not launch these and OpenAI responds directly. Is this the expected workflow?

I am trying to fix the automatic function-calling today.  When it works, plugins should automatically be invoked when they are relevant.

> _I had to correct some exceptions that were thrown when trying to use the chat. For example, the duplication of the chatid in the list._ Resolved using last commit.

If you encounter problems, please feel free to enter them in this PR so that I can address them.

> Thank you very much because thanks to this project I can understand much more about the functioning of SK.

My pleasure!

shared/ConfigurationBuilderExtensions.cs
```diff
@@ -0,0 +1,98 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
glahaye, line 1: File copied from the Kernel Memory project

shared/KernelMemoryBuilderExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
glahaye, line 1: File copied from the Kernel Memory project

shared/ServiceConfiguration.cs
```diff
@@ -0,0 +1,515 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
glahaye, line 1: File copied from the Kernel Memory project

webapi/Models/Response/ProposedPlan.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
```
glahaye, line 1: I'm thinking we could delete this file if we decided we don't need planning other than that provided by function calling

crickman (dismissed):

crickman (dismissed):

crickman (approved):

ezar:
> > @glahaye I have tried to use the Karna and Websercher plugins, but since the plan does not activate, they do not launch these and OpenAI responds directly. Is this the expected workflow?
> 
> I am trying to fix the automatic function-calling today. When it works, plugins should automatically be invoked when they are relevant.

I am trying to use the websearcher plugin that has the solution, but it seems that it is not able to invoke it. Before migrating to SK 1.4 using the planner, I made the request.
I have tried using the Klarna plugin and it seems to work.

glahaye:
@ezar I'll see what I can do about the websearcher plugin. I'm not officially given any time for this project anymore so it might take a bit :/

PR #824, opened 2024-02-29
Fix token budgeting

glahaye:
### Motivation and Context
OpenAI inserts a message under the hood for which we don't account, and thus throwing off our token budgeting. This causes us to sometimes send too many tokens when we approach a model's token limit. This results in errors.

### Description
Account for the OpenAI message inserted in our requests.

### Contribution Checklist

### DIFF
--- a/webapi/Plugins/Chat/ChatPlugin.cs
+++ b/webapi/Plugins/Chat/ChatPlugin.cs
@@ -637,7 +637,10 @@ public class ChatPlugin
     /// <returns>The remaining token limit.</returns>
     private int GetChatContextTokenLimit(ChatHistory promptTemplate, string userInput = "")
     {
-        const int ExtraOpenAiMessageTokens = 20; // OpenAI inserts a message under the hood that consumes some tokens
+        // OpenAI inserts a message under the hood:
+        // "content": "Assistant is a large language model.","role": "system"
+        // This burns just under 20 tokens which need to be accounted for.
+        const int ExtraOpenAiMessageTokens = 20;
 
         return this._promptOptions.CompletionTokenLimit
             - ExtraOpenAiMessageTokens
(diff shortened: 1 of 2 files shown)


webapi/Plugins/Chat/ChatPlugin.cs
```diff
@@ -636,7 +637,10 @@ private OpenAIPromptExecutionSettings CreateIntentCompletionSettings()
     /// <returns>The remaining token limit.</returns>
     private int GetChatContextTokenLimit(ChatHistory promptTemplate, string userInput = "")
     {
+        const int ExtraOpenAiMessageTokens = 20; // OpenAI inserts a message under the hood that consumes some tokens
```
glahaye, line 640: The fix

webapi/Plugins/Utils/TokenUtils.cs
```diff
@@ -121,14 +121,13 @@ internal static int TokenCount(string text)
...
-    /// e.g., "message": {"role":"assistant","content":"Yes }
+    /// e.g., "message": {"role":"assistant","content":"Yes" }
     /// </summary>
     /// <param name="authorRole">Author role of the message.</param>
     /// <param name="content">Content of the message.</param>
     internal static int GetContextMessageTokenCount(AuthorRole authorRole, string? content)
     {
-        var tokenCount = authorRole == AuthorRole.System ? TokenCount("\n") : 0;
-        return tokenCount + TokenCount($"role:{authorRole.Label}") + TokenCount($"content:{content}");
+        return TokenCount($"role:{authorRole.Label}") + TokenCount($"content:{content}\n");
```
glahaye, line 130: Conservative simplification

webapi/Plugins/Chat/ChatPlugin.cs
TaoChenOSU: Where does the 20 come from?

glahaye: Updated comment

TaoChenOSU (approved):

PR #839, opened 2024-03-06
Fix token budget + Clarify code

glahaye:
### Motivation and Context
The user message is not added to the GPT completion request when the token count of the prompt exceeds 50% of the configuration.

This fix is inspired by #836

### Description
Fix budget calculation and clarify / simplify code.

### Contribution Checklist

### DIFF
--- a/webapi/Plugins/Chat/ChatPlugin.cs
+++ b/webapi/Plugins/Chat/ChatPlugin.cs
@@ -262,6 +262,7 @@ public class ChatPlugin
 
         // Calculate max amount of tokens to use for memories
         int maxRequestTokenBudget = this.GetMaxRequestTokenBudget();
+        // Calculate tokens used so far: system instructions, audience extraction and user intent
         int tokensUsed = TokenUtils.GetContextMessagesTokenCount(metaPrompt);
         int chatMemoryTokenBudget = maxRequestTokenBudget
             - tokensUsed
@@ -387,7 +388,7 @@ public class ChatPlugin
         var result = await completionFunction.InvokeAsync(this._kernel, audienceContext, cancellationToken);
 
         // Get token usage from ChatCompletion result and add to original context
-        string? tokenUsage = TokenUtils.GetFunctionTokenUsage(result);
+        string? tokenUsage = TokenUtils.GetFunctionTokenUsage(result, this._logger);
         if (tokenUsage is not null)
         {
             context[TokenUtils.GetFunctionKey("SystemAudienceExtraction")] = tokenUsage;
@@ -433,7 +434,7 @@ public class ChatPlugin
         var result = await completionFunction.InvokeAsync(this._kernel, intentContext, cancellationToken);
 
         // Get token usage from ChatCompletion result and add to original context
-        string? tokenUsage = TokenUtils.GetFunctionTokenUsage(result);
+        string? tokenUsage = TokenUtils.GetFunctionTokenUsage(result, this._logger);
         if (tokenUsage is not null)
         {
             context[TokenUtils.GetFunctionKey("SystemIntentExtraction")] = tokenUsage;
--- a/webapi/Plugins/Chat/ChatPlugin.cs
+++ b/webapi/Plugins/Chat/ChatPlugin.cs
@@ -108,89 +108,6 @@ public class ChatPlugin
     }
 
     /// <summary>
-    /// Extract user intent from the conversation history.
-    /// </summary>
-    /// <param name="kernelArguments">The KernelArguments.</param>
-    /// <param name="cancellationToken">The cancellation token.</param>
-    private async Task<string> ExtractUserIntentAsync(KernelArguments kernelArguments, CancellationToken cancellationToken = default)
-    {
-        var tokenLimit = this._promptOptions.CompletionTokenLimit;
-        var historyTokenBudget =
-            tokenLimit -
-            this._promptOptions.ResponseTokenLimit -
-            TokenUtils.TokenCount(string.Join("\n", new string[]
-                {
-                    this._promptOptions.SystemDescription,
-                    this._promptOptions.SystemIntent,
-                    this._promptOptions.SystemIntentContinuation
-                })
-            );
-
-        // Clone the context to avoid modifying the original context variables.
-        KernelArguments intentExtractionContext = new(kernelArguments);
-        intentExtractionContext["tokenLimit"] = historyTokenBudget.ToString(new NumberFormatInfo());
-        intentExtractionContext["knowledgeCutoff"] = this._promptOptions.KnowledgeCutoffDate;
-
-        var completionFunction = this._kernel.CreateFunctionFromPrompt(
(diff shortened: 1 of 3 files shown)


webapi/Plugins/Chat/ChatPlugin.cs
```diff
@@ -107,89 +107,6 @@ public class ChatPlugin
         this._contentSafety = contentSafety;
     }
 
-    /// <summary>
```
glahaye, line 110: This code folder below, where it belongs

webapi/Plugins/Chat/ChatPlugin.cs
```diff
@@ -334,41 +251,41 @@ private async Task<CopilotChatMessage> GetChatResponseAsync(string chatId, strin
...
 
-        // Fill in the chat history with remaining token budget.
-        string allowedChatHistory = string.Empty;
-        var allowedChatHistoryTokenBudget = remainingTokenBudget - TokenUtils.GetContextMessageTokenCount(AuthorRole.System, memoryText);
-
-        // Append previous messages
+        // Add as many chat history messages to meta-prompt as the token budget will allow
         await this.UpdateBotResponseStatusOnClientAsync(chatId, "Extracting chat history", cancellationToken);
-        allowedChatHistory = await this.GetAllowedChatHistoryAsync(chatId, allowedChatHistoryTokenBudget, chatHistory, cancellationToken);
+        string allowedChatHistory = await this.GetAllowedChatHistoryAsync(chatId, maxRequestTokenBudget - tokensUsed, metaPrompt, cancellationToken);
```
glahaye, line 282: The fix

webapi/Plugins/Utils/TokenUtils.cs
```diff
@@ -62,50 +60,27 @@ public static class TokenUtils
     /// Gets the total token usage from a Chat or Text Completion result context and adds it as a variable to response context.
     /// </summary>
     /// <param name="result">Result context from chat model</param>
-    /// <param name="kernelArguments">Context maintained during response generation.</param>
-    /// <param name="logger">The logger instance to use for logging errors.</param>
-    /// <param name="functionName">Name of the function that invoked the chat completion.</param>
-    /// <returns> true if token usage is found in result context; otherwise, false.</returns>
-    internal static void GetFunctionTokenUsage(FunctionResult result, KernelArguments kernelArguments, ILogger logger, string? functionName = null)
+    /// <returns>String representation of number of tokens used by function (or null on error)</returns>
+    internal static string? GetFunctionTokenUsage(FunctionResult result)
```
glahaye, line 64: Simplified this method and put onus on caller to use its output instead of having obscure (but desired) side-effects

teresaqhoang: Why omit the log messages on error though? Can be helpful when debugging, esp for token calculation

webapi/Plugins/Chat/ChatPlugin.cs
```diff
@@ -334,41 +251,41 @@ private async Task<CopilotChatMessage> GetChatResponseAsync(string chatId, strin
...
         var userIntent = await AsyncUtils.SafeInvokeAsync(
             () => this.GetUserIntentAsync(chatContext, cancellationToken), nameof(GetUserIntentAsync));
-        chatHistory.AddSystemMessage(userIntent);
+        metaPrompt.AddSystemMessage(userIntent);
 
-        // Calculate the remaining token budget.
-        var remainingTokenBudget = this.GetChatContextTokenLimit(chatHistory, userMessage.ToFormattedString());
+        // Calculate max amount of tokens to use for memories
+        int maxRequestTokenBudget = this.GetMaxRequestTokenBudget();
+        int tokensUsed = TokenUtils.GetContextMessagesTokenCount(metaPrompt);
```
teresaqhoang, line 265: It's not intuitive what tokens are be calculated here. suggest adding a comment above

```suggestion
        // Calculate total tokens used so far: static system instructions, user intent, and audience extraction 
        int tokensUsed = TokenUtils.GetContextMessagesTokenCount(metaPrompt);
```

webapi/Plugins/Chat/ChatPlugin.cs
```diff
@@ -334,41 +251,41 @@ private async Task<CopilotChatMessage> GetChatResponseAsync(string chatId, strin
...
+        int chatMemoryTokenBudget = maxRequestTokenBudget
+            - tokensUsed
+            - TokenUtils.GetContextMessageTokenCount(AuthorRole.User, userMessage.ToFormattedString());
```
teresaqhoang, line 268: I think this would make more sense in the `GetMaxRequestTokenBudget` function since it always has to be accounted for. Here, it feels like it was just thrown in randomly and only accounted for when considering chat memory token budget (not the overall token budget)

or rename this variable to something like `remainingTokenBudget` to not tie it to chat memory

webapi/Plugins/Chat/ChatPlugin.cs
glahaye: Taken in

glahaye:
> And in the `ChatAsync` method, an `ArgumentException` is thrown in `SetSystemDescriptionAsync` because the chatId provided does not exists. I didn't have this issue before, so I'm not sure if this is related or if I should open an issue.

I don't believe this is related.

webapi/Plugins/Utils/TokenUtils.cs
glahaye: Logging put back in

webapi/Plugins/Chat/ChatPlugin.cs
glahaye: Actually that is precisely the problem: the user message should be considered ONLY for the memory budget but NOT for the overall budget.

The reason we DO want to to account for this in the memory budget is so that we ensure we later have room for the user message in our request.

The reason we DON'T want it in the overall budget is because it's the overall budget that's used to determine how many messages from the history we'll tack on to our request.

Remember that at this point, the user message has been saved to storage but **is not yet in the request** payload.

To illustrate how including the token count for the usage in the overall budget was a bug and that the fix is to keep it only for the memory budget, let's use an example.

Say you have a user message with a token count of 100.
Let's also say that the overall budget left after the system prompt, audience and user intent are tallied is 150.

Now, if you remove the user message token count (100) from the overall budget, you are left with only 50 tokens to use. Then, when we start adding chat history message to the request, we WON'T be able to add the most recent one (the user message with a size of 100) and we will send the request with no chat history messages whatsoever!

However, with this fix, we DON'T remove the user message token count (100) from the overall budget so we still have 150 tokens left. Then, when we add chat history messages to our request, we can add the most recent one (the user message with a size of 100) and up to an additional 50 tokens worth of messages.

teresaqhoang: I see I see, thanks for explaining in such detail!

teresaqhoang: Also great catch

teresaqhoang (approved):

PR #864, opened 2024-03-15
Refactor ChatController for Asynchronous Plugin Registration

Ahmed-Adel3:
This commit introduces asynchronous registration of plugins in the ChatController. The changes improve the efficiency of the chat function by allowing multiple plugins to be registered concurrently. This enhancement is expected to improve response times and overall performance of the chat functionality.

### Motivation and Context

1.	This change is required to improve the efficiency and performance of the chat functionality in the application.
2.	The current synchronous registration of plugins in the ChatController can lead to slower response times. This change solves this problem by introducing asynchronous registration, allowing multiple plugins to be registered concurrently.
3.	This contributes to the scenario where the application needs to handle multiple chat sessions simultaneously, each potentially requiring different sets of plugins. With the asynchronous registration, the application can now handle these scenarios more efficiently.
4.	This change does not directly fix an open issue, but it is a proactive measure to enhance the application's performance and scalability.

### Description
This edit is just a simple separation of registration of plugins [ JIRA - GRAPH - GITHUB - Custom plugins ] each one of them in a separate **Asynchronous** function and registering the enabled plugins simultaneously, as this process occurs each time a message is sent to chatbot.

### Contribution Checklist


Ahmed-Adel3:
@microsoft-github-policy-service agree

webapi/Controllers/ChatController.cs
```diff
@@ -194,87 +196,112 @@ private async Task RegisterFunctionsAsync(Kernel kernel, Dictionary<string, stri
...
+
+    private async Task RegisterJiraPlugin(Kernel kernel, string JiraAuthHeader, KernelArguments variables)
+    {
+        this._logger.LogInformation("Registering Jira plugin");
+        var authenticationProvider = new BasicAuthenticationProvider(() => { return Task.FromResult(JiraAuthHeader); });
+        var hasServerUrlOverride = variables.TryGetValue("jira-server-url", out object? serverUrlOverride);
+
+        await kernel.ImportPluginFromOpenApiAsync(
+            pluginName: "JiraPlugin",
+            filePath: Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!, "Plugins", "OpenApi/JiraPlugin/openapi.json"),
```
glahaye, line 248: Can't this use _pluginDirectoryPath?

webapi/Controllers/ChatController.cs
```diff
@@ -194,87 +196,112 @@ private async Task RegisterFunctionsAsync(Kernel kernel, Dictionary<string, stri
...
+        BearerAuthenticationProvider authenticationProvider = new(() => Task.FromResult(GraphAuthHeader));
+        GraphServiceClient graphServiceClient = this.CreateGraphServiceClient(authenticationProvider.GraphClientAuthenticateRequestAsync);
+
+        kernel.ImportPluginFromObject(new TaskListPlugin(new MicrosoftToDoConnector(graphServiceClient)), "todo");
+        kernel.ImportPluginFromObject(new CalendarPlugin(new OutlookCalendarConnector(graphServiceClient)), "calendar");
+        kernel.ImportPluginFromObject(new EmailPlugin(new OutlookMailConnector(graphServiceClient)), "email");
+        return Task.CompletedTask;
+    }
+
+    private async Task RegisterCustomPlugins(Kernel kernel, object? customPluginsString, Dictionary<string, string> authHeaders)
```
glahaye, line 268: Since we might be registering more than one custom plugin, it seems to me that it would make more sense to have this method return a IEnumerable of Tasks (perhaps a List) and then do tasks.AddRange() so that these custom plugins can be loaded in parallel too.

For example:

tasks.AddRange(RegisterCustomPlugins(...));

webapi/Controllers/ChatController.cs
Ahmed-Adel3: It's updated by by introducing a new function "GetPluginFullPath" to get the filePath depending on the path of openapi.json which is passed as a parameter to the function
And so "_pluginDirectoryPath" is completely removed

webapi/Controllers/ChatController.cs
Ahmed-Adel3: RegisterCustomPlugins is now returning IEnumerable<Task> and yield returning the result of kernel.ImportPluginFromOpenAIAsync

glahaye (approved):

Issue #718, opened 2023-12-13
Considerably slow responses obtaining threads at launch


UTMark:
Good morning everyone!

I would like to report a bug that could make the application feel unusable if it were deployed in a production environment. The issue is in two parts but overall what it amounts to is it can take a considerable amount of time to wait for the chat application to load due to the extreme slow responses for each chat thread as they are loaded in.

Example on how long chat messages take to load on first entry: (4-9s)
![image](https://github.com/microsoft/chat-copilot/assets/153741058/a0bbc942-05f2-4328-b19e-3ec9b8e8c833)

Now imagine if you had 10 threads or 100 threads:
![image](https://github.com/microsoft/chat-copilot/assets/153741058/32bae481-c407-4bb2-bc95-333c06fd4859)

---

**Here are some details about the server:**

Copilot Build Version: v0.9
SKU: PremiumV3 (P2v3)
CosmosDB: 10K RU /each
 - Index usage 24.4kb (monitor)
 - 384MB Size
 - 10K RU allocation per table, but a tiny fraction is being used right now
Memory: AzureCognitiveSearch

---

**Cosmos Index Policy:**
`{
    "indexingMode": "consistent",
    "automatic": true,
    "includedPaths": [
        {
            "path": "/*"
        }
    ],
    "excludedPaths": [
        {
            "path": "/\"_etag\"/?"
        }
    ]
}`

**Thoughts:**

I'm not a CosmosDB expert so I'm not sure why the little turtle is showing between 4s - 9s per request for each thread. I am more than willing to investigate with a bit of guidance on what to look into. If this were SQL, I would have imagined it would have been an indexing issue which is why I provided the indexing policy. 

But in hindsight, I was wondering that if this is a bug and has a resolution to it, then why not obtain the threads in bulk and limit it to a certain maximum number and paginate as the user scrolls down the historical items? I will continue to look into this issue myself but some of these technologies are new to me. Thank you for your time!

UTMark:
I think what may be happening is that the cosmos database may need to have hierarchical partitions for the containers. For example:

**chatmessages:** id /chatId /userId
**chatparticipants:** id /userId /chatId

Partition might also need to be considered?

[...]

PR #902, opened 2024-03-27
Respect order, number to skip and number to take of chat messages at underlying DB level rather than higher

glahaye:
### Motivation and Context
As described in #718, we load ALL the messages from ALL of the chats of the user in the frontend at log in.

Also, no matter how many messages we want, we read them ALL at the DB level and then throw some away if the number read is too higher at the service level.

### Description
Now, with this change, we actually respect the order, the number to skip and the number to take of chat messages at the underlying DB level rather than at higher service level.

This enables us to save some DB activity and make our queries from the frontend eventually a lot tighter,

This change was made in a simple manner which doesn't change the underlying architecture.

### Contribution Checklist


webapi/Controllers/ChatHistoryController.cs
```diff
@@ -183,19 +183,15 @@ public async Task<IActionResult> GetAllChatSessionsAsync()
     [Authorize(Policy = AuthPolicyName.RequireChatParticipant)]
     public async Task<IActionResult> GetChatMessagesAsync(
         [FromRoute] Guid chatId,
-        [FromQuery] int startIdx = 0,
+        [FromQuery] int skip = 0,
```
crickman, line 186: `skip` may be problematic for a growing list of messages in descending order (multiuser).  I've seen this commonly expressed as the identifier of the last message in the previous selection.  I'm not sure that's so impactful here.

crickman (approved):

webapi/Controllers/ChatHistoryController.cs
glahaye: I totally understand what you are saying. I was tempted to remove skip altogether. It is not currently used nor do I anticipate it ever to be.

The "proper" way of designing this endpoint would probably be to have it support OData syntax instead of the current half-baked API we have here.

But given that we're merely trying to keep the lights on here and not inject too much work to this project, I wanted to at least:
- stop fetching ALL entries and then throwing all those over the number of messages we truly want
- get the LATEST messages when requesting fewer than exist in a chat so that we can populate the "meta-prompt" with the latest data instead of suddenly having a hole in our chart where we only have old messages and newer ones are not returned

PR #994, opened 2024-05-29
fix(webapp): Fix 989 issue

william-daconceicao:
### Motivation and Context

Allow the project to work anonymously

### Description

Fix 'getLogger' error that prevent the webapp to start anonymously.


william-daconceicao:
@microsoft-github-policy-service agree

glahaye:
Thanks for the fix!

PR #1203, opened 2024-11-08
Fix typos, docs, and code style

dluc:
* Fix typos
* Fix configuration docs, e.g. use correct section names
* Fix code style

### DIFF
--- a/webapi/Controllers/PluginController.cs
+++ b/webapi/Controllers/PluginController.cs
@@ -48,7 +48,7 @@ public class PluginController : ControllerBase
     /// <param name="manifestDomain">The domain of the manifest.</param>
     /// <returns>The plugin's manifest JSON.</returns>
     [HttpGet]
-    [Route("pluginManifests")] // TODO: Fix name and test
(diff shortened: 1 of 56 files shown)


webapi/Controllers/PluginController.cs
```diff
@@ -48,9 +48,9 @@ public PluginController(
     /// <param name="manifestDomain">The domain of the manifest.</param>
     /// <returns>The plugin's manifest JSON.</returns>
     [HttpGet]
-    [Route("pluginManifests")]
+    [Route("pluginManifests")] // TODO: Fix name and test
```
TaoChenOSU, line 51: Question: is the TODO intended?

TaoChenOSU (dismissed):

webapi/Controllers/PluginController.cs
dluc: yes, I believe the name should be singular

dluc, line 51:
```suggestion
    [Route("pluginManifests")] // TODO: Fix name and test - use singular?
```

